PR #49, opened 2017-12-01
Fix rotation of input winds: Cassini to earth relative

davegill:
Modify the existing met_to_map routine to use the correct rotation
angle, alpha. Previously, the rotation angle was computed with the lat,lon
for the d(lat)/d(y) and d(lon)/d(y) from the interpolated data on the
WRF domain. The correct lat and lon for these d/dy values is to use the
lat lon from the original input data.

The d(lat)/d(y) and d(lon)/d(y) computations use the existing latlon_to_ij
and the inverse ij_to_latlon in the module_map_utils.F file. This module is
now required to be "used" in rotate_winds_module.F, and the associated
Makefile for metgrid is also modified to include this new dependency.

The changes are only made for the Cassini projection (no other test data
exists to try out the other input projections). However, it looks like the
rotation angle alpha needs to have a -1.0 scale factor. Basically, if alpha
is the angle to rotate to go from Earth relative to a projection, then -alpha
is the required angle to take the winds from the projection (in this case, 
Cassini) back to Earth relative.

Tests conducted: 

1. GFS input vs RAP (rotated lat lon) input, each with 4 geogrid 
projections: Lambert, Polar, Mercator, Regional lat/lon. A global domain
is not possible for the regional RAP input data.
2. The sensitivity of the computation of d(lon)/dy and d(lat(dy) was 
considered (dy = 1 is one grid cell unit in the input data). Using dy = 
0.1 vs 1.0 resulted in rotation differences < 0.1 degrees (possibly this
is small due to the input RAP data being pretty high horizontal resolution).

### DIFF
--- a/metgrid/src/Makefile
+++ b/metgrid/src/Makefile
@@ -66,7 +66,7 @@ queue_module.o: module_debug.o
 
 read_met_module.o: constants_module.o misc_definitions_module.o module_debug.o met_data_module.o
 
-rotate_winds_module.o: bitarray_module.o constants_module.o llxy_module.o misc_definitions_module.o module_debug.o
+rotate_winds_module.o: bitarray_module.o constants_module.o llxy_module.o misc_definitions_module.o module_debug.o module_map_utils.o
 
 storage_module.o: datatype_module.o minheap_module.o misc_definitions_module.o module_debug.o parallel_module.o module_stringutil.o
 
--- a/metgrid/src/rotate_winds_module.F
+++ b/metgrid/src/rotate_winds_module.F
@@ -161,7 +161,6 @@ module rotate_winds_module
             ! Rotate U field
             do j=us2,ue2
                do i=us1,ue1
-
                   diff = idir * (xlon_u(i,j) - proj_stack(current_nest_number)%stdlon)
                   if (diff > 180.) then
                      diff = diff - 360.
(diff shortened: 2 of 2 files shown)


davegill:
@weiwangncar @smileMchen @mgduda 
From the figures, the rotated lat-lon works for several projections. With these mods, the regular lat-lon (such as GFS) works for the several WRF projections.

Is there an issue with either the results or with the software modification? I know it will take everyone a bit to look back over this.

weiwangncar:
I'm ok with the change. Perhaps Ming can do a timing test to see if the timing changes much for the other projections not affected by the change.

smileMchen:
I have run a few tests using the original codes and the modified codes. Lambert, Mercator, and Polar Stereographic map projections are tested. Grid number is 440 x 240. The results show that for all the types of map projection, the running time that is taken for processing 4 times of GFS data is similar (actually almost the same)  using  original and modified codes. 
I think we can put the modified code (rotate_winds_module.F) to WPS repository, which will enable people to process HRRR data.

mgduda:
@smileMchen I understand from @davegill that you're going to run a few other tests to ensure that these changes don't break other rotated projections. When you're satisfied that this PR is ready to merge, can you add your review and remove the "Don't merge yet!" label?

smileMchen:
We process NAM and GFS data using the modified code. Below is the figure showing the wind vectors at 700 hPa, which indicates that NAM and GFS are quite similar and the code works fine. We also process NAM data using codes with and without modification. The results are identical (Figure not shown). The above tests  suggest that the modified codes are in good shape.  I have no objection to merge them into the master. 
dave_nam_gfs1

Issue #72, opened 2018-10-01
Need additional logic to detect WRF code downloaded as GitHub archive


mgduda:
If users download the WRF source code as a GitHub "archive" file from https://github.com/wrf-model/WRF/releases , the unpacked source directory will not be named `WRF`, but will have a name like `WRF-4.0`.

Somehow, we need to be able to find this WRF directory when configuring the WPS so that the path to the WRF I/O API can be added in the WPS `configure.wps` file.

Suggestions that have been offered:
* Have the user type in the path to the WRF source code (or the name of the directory) when running the WPS `configure` script

mgduda:
A little progress has been made with PR #102 , but a general solution has yet to be implemented.

PR #73, opened 2018-10-02
Add additional logic in arch/preamble for finding WRF source directory

mgduda:
Add additional logic in `arch/preamble` for finding the WRF source directory

The make logic in `arch/preamble` for setting the `WRF_DIR` variable to the location
of the compiled WRF source code previously looked either for a directory named
`../WRF` or a directory named `../WRFV3`. In anticipation of users downloading
the WRF code as a GitHub archive file, which will unpack to a name like
`WRF-4.0.1`, we need more sophisticated logic for finding this directory.

This commit adds checks for the WRF code in `../WRF-4.0.1` and `../WRF-4.0`.
In future we will likely need a more general solution (e.g., to handle the
eventual WRF 4.1 release), but as an immediate fix, checking for the WRF
4.0 or 4.0.1 releases should be sufficient.

### DIFF
--- a/arch/preamble
+++ b/arch/preamble
@@ -33,14 +33,26 @@ PERL			=	perl
 RANLIB			=	echo
 
 #
-# Set path to compiled WRF code, which is assumed to be ../WRF for v4.x, or ../WRFV3 for v3.x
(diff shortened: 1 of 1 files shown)


mgduda:
This PR addresses Issue #72 .

Issue #71, opened 2018-09-07
cheyenne new paths to static data


davegill:
Top of the repo geogrid error message on cheyenne:
```
ERROR: Could not open /glade/p/work/wrfhelp/WPS_GEOG/topo_gmted2010_30s/index
```

We need to remove the "p" subdirectory from namelist.wps
```
geog_data_path = '/glade/work/wrfhelp/WPS_GEOG/'
```

mgduda:
PR #75 addresses this issue.

PR #75, opened 2018-10-02
Update default path to geog data on /glade filesystem on NWSC Cheyenne

mgduda:
Update default path to geog data on `/glade` filesystem on NWSC Cheyenne

The default path to the static geographical data previously pointed to
the `/glade/p/work` filesystem on the NWSC Cheyenne cluster. The `/glade` filesystem
was recently restructured so that `/glade/p/work` no longer exists.

Now, the geographical datasets are available in `/glade/work/wrfhelp/WPS_GEOG`.

### DIFF
--- a/namelist.wps
+++ b/namelist.wps
@@ -36,7 +36,7 @@
  truelat1  =  30.0,
  truelat2  =  60.0,
  stand_lon = -98.0,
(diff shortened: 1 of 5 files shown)


mgduda:
This PR addresses Issue #71 .

PR #80, opened 2018-10-18
Cray: compack.f, Grib2 fixed vs free format, illegal use of "&"

davegill:
This is a fixed format file. There was a trailing "&" and no acceptable character in column 6 of
the continuing line. 

The trailing "&" was removed, and column six on the continuing line now sports a spiffy "&".

With the Cray ftn compiler, without this mod, the code does not compile. When using this mod,
this routine compiles. Other compilers let this slide through the cracks.


davegill:
Better method is to remove this file from being explicitly mentioned in the g2 Makefile. That means the .a is not dependent on a broken .o file. This keeps compatibility with NCEP issued g2 libraries.

PR #81, opened 2018-10-18
Cray: Grib2 pack_gp.f fixed vs free format "&" mod

davegill:
This one is on Cray and their ftn Fortran compiler.

The fix involves removing a sequence of trailing "&" characters for the declaration of an array of
integer parameters. 

This is a fixed format file, so anything in column 72 and beyond should be ignored. The "&" character is in column 73 (in an attempt to make the code compliant with f77 or modern Fortran). This should
compile OK, but does not.


davegill:
Better way to handle this is to remove this file from the local Makefile. That keeps g2 lib compatibility and also allows Cray to successfully build the .a file.

PR #82, opened 2018-10-18
Cray mods: rrpr.F needs write statement clean-up

davegill:
This is a fix for the Cray ftn compiler.

This is a free-format file. The proper syntax when splitting a string across more than one line is
to end each line with an "&" character and to also start the next line with an "&" character.

For example, we want this:
long_char = 'This is a lon&
                     &g string across mult&
                     &iple lines'

The original code did not include the leading "&" character on the following line.


davegill:
Jordan Powers doing this PR

PR #83, opened 2018-10-23
Bugfix to support Cray compiler: fixed-format modification.

powersjg:
Continuation character at end of line in original code
was not permitted by Cray compiler.  Removed "&" to support
fixed-format Fortran in subroutine compack.f.

### DIFF
--- a/ungrib/src/ngl/g2/compack.f
+++ b/ungrib/src/ngl/g2/compack.f
@@ -254,7 +254,7 @@
      &                  kbit,novref,lbitref,ier)
            if(ier/=0) then
               ! Dr. Glahn's algorithm failed; use simple packing method instead.
- 1099         format('G2: fall back to simple algorithm (glahn ier=',I0,&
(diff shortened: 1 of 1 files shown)


powersjg:
@mgduda @jimbresch 

HI Michael and Jim:
Made a mod to an Ungrib routine support the Cray compiler.  Mod is in compack.f.
Trying out the git process.  Dave has these PRs, too, but we'll only merge my versions and not Dave's versions.  There will be two other similar, small mods for the Cray compiler coming. 

Best,
Jordan

jimbresch:
I disagree with the change. The ngl directory contains the NCEP Grib Libraries. Making additional, WPS-specific changes to these routines adds an unnecessary maintenance burden. It means more work when we update these libraries with a newer version from NCEP.
More importantly, the routines compack and pack_gp are packing routines and not used by WPS at all. A better solution is to simply remove compack and pack_gp from the OBJS line in ngl/g2/Makefile. I've tested that and it works fine. We can leave the source code in the directory, it just won't be compiled.

PR #84, opened 2018-10-23
Mods to support Cray compiler: fixed-format modification.

powersjg:
Continuation characters at end of lines in original code
were not permitted by Cray compiler.  Removed multiple "&"s to
support fixed-format Fortran in subroutine pack_gp.f.


powersjg:
@mgduda @jimbresch

Michael and Jim: 
This is the second of a trio of Ungrib fixes for the Cray compiler.

-JGP

powersjg:
This PR is being rescinded as there is better way to address the issue.  
The target routine is actually not necessary, and thus the routine itself 
will be targeted for removal. The makefile will also be adjusted. These 
mods will be proposed in another PR.

PR #100, opened 2019-01-18
Remove -f90= and -cc= options from MPI wrappers

mgduda:
This merge corrects build warnings and errors caused by the use of `-f90=`
and `-cc=` options with the MPI compiler wrappers. Examples include:
```
gfortran: error: unrecognized command line option ‘-f90=gfortran’

ifort: command line warning #10006: ignoring unknown option '-f90=ifort'

pgfortran-Error-Unknown switch: -f90=pgf90
```

Depending on the MPI implementation, it appear that the mpif90 and mpicc
wrappers may or may not pass the `-f90=` and `-cc=` options directly to
the underlying compiler, where these options may generate either warnings
or errors.

The original intent of using, e.g., `mpif90 -f90=gfortran` was to permit
this use of a different Fortran compiler than the default selected by
the MPI wrapper. However, this appears to be not widely supported, and
it may be preferable to just remove these options and to assume that
the default compilers selected by the MPI wrappers are correct.

Note: The `-f90=` and `-cc=` options have not been removed from stanzas that
will likely be deprecated in future, either because the hardware is no longer
produced (PowerPC Apple hardware) or the compiler is no longer being
developed (g95 and PathScale).

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -61,8 +61,8 @@ COMPRESSION_INC     = CONFIGURE_COMP_I
 FDEFS               = CONFIGURE_FDEFS
 SFC                 = ifort
 SCC                 = icc
-DM_FC               = mpif90 -f90=ifort
-DM_CC               = mpicc -cc=icc
+DM_FC               = mpif90
+DM_CC               = mpicc
 FC                  = CONFIGURE_FC
 CC                  = CONFIGURE_CC 
 LD                  = $(FC)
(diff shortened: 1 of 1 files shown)


mgduda:
@llpcarson Would you be able to check whether the changes in this PR have any adverse consequences for building on NOAA machines?

llpcarson:
Michael - we've been directed to not use NOAA HPC resources during the
shutdown, unfortunately.  I'd suggest that this PR go in, and we can
fix/update later if need be.

PR #102, opened 2019-01-21
Add extensible logic for finding WRF build, e.g., in ../WRF-4.0.3

mgduda:
This merge replaces the logic in `arch/preamble` for locating compiled WRF source
code with more extensible logic in the `configure` script.

The path to the compiled WRF source can now be specified by the environment
variable `$WRF_DIR`, or an attempt will be made to find the compiled source in
`../WRF-4.0.3`, `../WRF-4.0.2`, ..., `WRFV3`.

### DIFF
--- a/configure
+++ b/configure
@@ -150,8 +150,47 @@ if [ $? -eq 0 ] ; then
     fi
 fi
 
+wrf_dir="none"
+standard_wrf_dirs="WRF WRF-4.0.3 WRF-4.0.2 WRF-4.0.1 WRF-4.0 WRFV3"
+
+#
+# If no WRF_DIR environment variable is set, try to locate a WRF build in one
+# of the expected directory names one directory level up; otherwise, try to use
+# the WRF I/O library from the code in $WRF_DIR
+#
+if [ -z "$WRF_DIR" ]; then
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -150,8 +150,43 @@ if [ $? -eq 0 ] ; then
...
+      exit
+   fi
+
+   echo "Using WRF I/O library in WRF build identified by \$WRF_DIR: ${WRF_DIR}"
+   wrf_dir=$WRF_DIR
+fi
+
+if [ $wrf_dir == "none" ]; then
+   echo ""
+   echo " No compiled WRF code found... Check that WRF has been compiled or whatever"
```
kkeene44: @mgduda 
Did you mean to leave this as "....or whatever?"

mgduda: That was a test to see how closely anyone checks these PRs. A+. I'll push an amended commit that improves that message.

mgduda: I've just pushed an amended commit. Hopefully the message is more informative now.

kkeene44:
@mgduda 
I accidentally tested out probably every scenario for this and everything seems to work great. I like this concept. I'll approve it.

Issue #103, opened 2019-01-22
Compilation warnings about truncated string initialization


mgduda:
With the gfortran compiler, there are quite a few warnings about trunctated string
initialization that are generated during compilation of the `g1print.exe` utility:
```
gfortran -c -ffree-form -O -fconvert=big-endian -frecord-marker=4 g1print.f90
g1print.f90:664:71:

        'VSTR','M FLX','LMH','LMV','SGLYR','NLAT','ELON','UMAS','VMAS','XPRATE',&
                                                                       1
Warning: Initialization string starting at (1) was truncated to fit the variable (5/6)
```

mgduda:
This issue was corrected by PR #104 and can be closed now.

PR #104, opened 2019-01-24
Increase length of field names in GRIB tables in g1print from 5 to 7 chars

mgduda:
This merge increase the length of field names in GRIB tables in the g1print
utility to 7 characters (from a previous length of 5 characters).

In order to avoid compilation warnings about truncated strings (and to prevent
strings from being truncated), increase the length of field names in the GRIB
tables built in to the g1print utility from 5 to 7 characters.

This merge also adds CLWC and CIWC to the ECMWF GRIB table.

This merge addresses Issue #103 .

### DIFF
--- a/ungrib/src/g1print.F
+++ b/ungrib/src/g1print.F
@@ -3,6 +3,7 @@
 !  Usage: "g1print [-v] [-V] filename"
 !
 !  Originally from gribscan.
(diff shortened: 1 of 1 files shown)


mgduda:
I can confirm that the compilation warnings from `gfortran` are gone, and the `g1print` utility seems to be working as expected with ERA-I and FNL GRIB1 files.

PR #107, opened 2019-02-15
Change Vtable.GFS so the default sea level pressure is MSLET

davegill:
Replaces PR #97 "Change Vtable.GFS so the default sea level pressure is MSLET - BAD HISTORY"

NCEP has included the MSLET sea level pressure field in its pressure-level
grib files since May 10, 2011. This is in addition to the PRMSL sea level
pressure field (the previous default SLP in WPS). The PRMSL field is
smoothed/truncated. For the lower horizontal resolutions of GFS output in
years past this wasn't a significant issue. For the current higher
resolutions, the smoothing is very noticeable. (See attached figures).
SLP may be used to compute surface pressure in real. WRF tests with this
change show only minor differences. The main impact of using MSLET instead
of PRMSL is if metgrid output is used for plotting or verification.

Impact on the SLP field outside the areas of tropical storms:
Any system with a small-scale minimum or maximum will be smoothed away in
PRMSL. Practically, only cyclones (tropical or extratropical) have a
noticeable difference with the new field.

The only file changed is the GFS Vtable. The new SLP field is provided and the old
SLP field is deleted. A message is provided to users on how to get the original SLP
field.

Running the two different Vtable.GFS files (before vs after mods), the only field that
changes is PMSL.

The smoothed field (top figure) is PRMSL which is the current default.
The bottom figure is MSLET the proposed default SLP field in WPS.

SMOOTHED SLP
met2

UN-SMOOTHED SLP
met3

### DIFF
--- a/ungrib/Variable_Tables/Vtable.GFS
+++ b/ungrib/Variable_Tables/Vtable.GFS
@@ -11,7 +11,7 @@ Param| Type |Level1|Level2| Name     | Units   | Description                    
   33 | 105  |  10  |      | UU       | m s-1   | U                 at 10 m               |  0  |  2  |  2  | 103 |
   34 | 105  |  10  |      | VV       | m s-1   | V                 at 10 m               |  0  |  2  |  3  | 103 |
    1 |   1  |   0  |      | PSFC     | Pa      | Surface Pressure                        |  0  |  3  |  0  |   1 |
-   2 | 102  |   0  |      | PMSL     | Pa      | Sea-level Pressure                      |  0  |  3  |  1  | 101 |
+ 130 | 102  |   0  |      | PMSL     | Pa      | Sea-level Pressure                      |  0  |  3  | 192 | 101 |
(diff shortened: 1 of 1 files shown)


weiwangncar:
Do we have absolute confidence that this will not break anything? I'd like to see this old MSLP retained somehow in the Vtable, just in case that someone has to use the old one. We can release it as part of the alpha release, and revisit this before the final one.

davegill:
@weiwangncar 
Wei,
I ran a 24-h forecast and looked at the initial SLP (using today's GFS). The "unsmoothed" is noisier, as you would expect. 

Jim's PR has the information in it (the ver last line of the Vtable.GFS) to allow users to go back to the original SLP field. 
```
#  2 | 102  |   0  |      | PMSL     | Pa      | Sea-level Pressure                      |  0  |  3  |  1  | 101 |
```

As you said, we can modify this over the next 6 weeks. We could even have different Vtable.GFS files.

PR #108, opened 2019-02-15
Merge master into develop, pick up v4.0.3 updates, prior to v4.1 merge

davegill:


mgduda:
@davegill @kkeene44 We should chat a bit next week about procedure. Generally, merging `develop` from someone's fork onto the main `develop` branch is not how I think we should be merging `master` into `develop`.

PR #116, opened 2019-03-04
Add LOGSFP entry to METGRID.TBL.ARW

mgduda:
Model-level ECMWF datasets typically provide a logarithm of the surface
pressure, which may be used by the calc_ecmwf_p utility to produce a full 3-d
pressure field. In order to silence warnings from metgrid when the LOGSFP field
is found in intermediate files, this commit adds an entry to the METGRID.TBL.ARW
file for this field. Since the LOGSFP field is not used by WRF, the "output=no"
specification is also added to the LOGSFP entry.

### DIFF
--- a/metgrid/METGRID.TBL.ARW
+++ b/metgrid/METGRID.TBL.ARW
@@ -1034,3 +1034,9 @@ mpas_name=tslb
         masked=water
         fill_missing=1.
 ========================================
+# Typically from ECMWF model-level data and used by calc_ecmwf_p.exe
+# Not needed by WRF
+name=LOGSFP
+        output=no
+        interp_option=four_pt
+========================================


mgduda:
@weiwangncar I've targeted the `release-v4.1` branch with this PR, rather than the `develop` branch.

weiwangncar:
Shall we worry about the other METGRID.TBLs?

weiwangncar:
Shall we consider other METGRID.TBLs?

mgduda:
I'm disinclined to modify the NMM METGRID.TBL file, and I'm not actually sure who uses the `AFWA`, `rap`, and `ruc` ARW METGRID.TBL files. Given that this is a PR to improve the WPS experience for the typical user, and under the assumption that the vast majority of users will be using the default `METGRID.TBL.ARW` file, it's not unreasonable to only modify that file in my opinion.

llpcarson:
No need to update the NMM table at this time for this field.

weiwangncar:
I'm ok with it.

jimbresch:
Well, I disapprove of this pull request. This warning message is telling a user that the default interpolation settings will be used for this field. The warning served its purpose in notifying an ignorant user about how logsfp is supposed to be used by the WRF system. A savvy user would know to ignore this warning just as real will ignore the logsfp field.

I often use ungrib/metgrid to decode and plot fields from grib files on my wrf grid that aren't used by wrf. These default warning messages are common and harmless. If you change the METGRID.TBL to output=no, then you wouldn't be able to plot the logsfp without modifying the table. This should not be the default behavior.

If you want to add an additional message (e.g. 'LOGSFP will not be used to initialize WRF') or perhaps change the angry 'WARNING' to a more gentle 'NOTE' in the message, I can agree to that.

mgduda:
@jimbresch I think the argument for this PR goes something like the following: When users initialize a model simulation using model-level ECMWF data, they're doing something that is completely valid and supported, and in that case, we should minimize the number of warning messages that they see.

You've proposed to change the WARNING message to a NOTE, which is not a bad idea, but I'd propose that the above argument still holds, and that we should minimize the number of harmless messages that a user sees. Harmless messages either cause needless concern among new users, or condition users to ignore messages from the WRF modeling system.

As you said, we could change the message to something like "LOGSFP will not be used to initialize WRF", but that could be a misleading message in cases where WRF developers have added a new input field in WRF's Registry file. And, by implication of omission, any time this message does not appear, users may well believe that all fields in their intermediate files will be used in the WRF simulation, which may not always a valid assumption.

We could omit the `output=no` part of the LOGSFP entry, but my guess is that the number of users who want to plot LOGSFP from the metgrid output files is almost nil.

I would certainly oppose the idea of having metgrid neither mention in a log message nor output all unrecognized fields in its input intermediate files, but in the case of LOGSFP, which is used for a very specific purpose prior to metgrid (i.e., to derive a full 3-d PRESSURE) field, but never after the metgrid stage, I think this addition to the METGRID.TBL file is reasonable.

jimbresch:
OK, essentially what I object to here is using METGRID.TBL to filter out fields that have been properly requested via the Vtable. Basically, you're using 'output=no' in the METGRID.TBL to make up for deficiencies in the calc_ecmwf_p.F program. Wouldn't it be pretty simple to the change calc_ecmwf_p.F to read and write the fields it finds in the intermediate input file except for LOGSFP? Then, only one intermediate file (without LOGSFP) is passed on to metgrid.

mgduda:
I agree that it would be ideal for `calc_ecmwf_p` to only write out 3-d pressure and not LOGSFP. The difficulty is that LOGSFP resides in a file created by `ungrib` and this file is not rewritten by `calc_ecmwf_p`. `calc_ecmwf_p` only writes a new intermediate file with 3-d pressure, RH, and GHT; so, we'd need to make a non-trivial change to `calc_ecmwf_p` to rewrite all of its input intermediate files, which would at best lead to a lot of extra disk I/O.

I wouldn't be opposed to removing the `output=no` option from the proposed LOGSFP entry. The important point in my opinion is that we add some entry to the `METGRID.TBL` file to handle LOGSFP, rather than letting the field be treated with the default nearest-neighbor interpolation and producing a warning message that may confuse users needlessly.

davegill:
@mgduda @jimbresch @weiwangncar 
Michael and Jim,
There are few default fields that move from ungrib to metgrid that will not be sent into the real program. The LOGSFP field is special:
1. It is only arrives at metgrid from in an input file from ungrib because we don't remove unnecessary files in our calc_ecmwf processing.
2. The LOGSFP field is identified and known to not be used by real or WRF, so it does not need to go downstream.
3. For a first time user of EC data, a message about unknown files is going to be cause concern about things running correctly.
4. Putting a mod in a table, as opposed to compiled code, to handle preferred default behavior for an indicated variable seems reasonable.

If a user wants to output a field that is given "output=no", a message in metgrid stating that a certain field will not be output based on the table defaults could be provided.

mgduda:
We can consider adding messages to the metgrid output for a future release, but it's rather late in the v4.1 release cycle to be making these sorts of changes. There are other examples of fields in the `METGRID.TBL` file with `output=no`, e.g., `HGTT`, so I don't see why this should be problematic for `LOGSFP`.

PR #119, opened 2019-03-19
Fix mismatch integer kind in IAND intrinsic calls in g2 library

mgduda:
The calls to the IAND intrinsic in the g2 library's ilog2_2 and ilog2_1
function used arguments of 'i' and 'i-1', where 'i' is of kind=2 or kind=1.
The XL compiler fails to compile this code with the complaint that
the arguments to IAND are of mismatched kind. It appears that 'i-1' is
being promoted to the default integer kind, unless we declare the constant
with _2 or _1.

### DIFF
--- a/ungrib/src/ngl/g2/intmath.f
+++ b/ungrib/src/ngl/g2/intmath.f
@@ -169,7 +169,7 @@
       ilog2_2=0
       i=i_in
       if(i<=0) return
-      if(iand(i,i-1)/=0) then
+      if(iand(i,i-1_2)/=0) then
          !write(0,*) 'iand i-1'
          ilog2_2=1
       endif
@@ -204,7 +204,7 @@
       ilog2_1=0
       i=i_in
       if(i<=0) return
-      if(iand(i,i-1)/=0) then
+      if(iand(i,i-1_1)/=0) then
          !write(0,*) 'iand i-1'
          ilog2_1=1
       endif
--- a/ungrib/src/ngl/g2/intmath.f
+++ b/ungrib/src/ngl/g2/intmath.f
@@ -169,6 +169,8 @@
       ilog2_2=0
       i=i_in
       if(i<=0) return
+! WPS modification for the XL compiler
+!      if(iand(i,i-1)/=0) then
       if(iand(i,i-1_2)/=0) then
          !write(0,*) 'iand i-1'
          ilog2_2=1
@@ -204,6 +206,8 @@
       ilog2_1=0
       i=i_in
       if(i<=0) return
+! WPS modification for the XL compiler
+!      if(iand(i,i-1)/=0) then
       if(iand(i,i-1_1)/=0) then
          !write(0,*) 'iand i-1'
          ilog2_1=1


ungrib/src/ngl/g2/intmath.f
```diff
@@ -204,7 +204,7 @@ function ilog2_1(i_in)
       ilog2_1=0
       i=i_in
       if(i<=0) return
-      if(iand(i,i-1)/=0) then
+      if(iand(i,i-1_1)/=0) then
          !write(0,*) 'iand i-1'
          ilog2_1=1
```
davegill: Similarly here, do we say
```
ilog2_1=1_1
```

ungrib/src/ngl/g2/intmath.f
```diff
@@ -169,7 +169,7 @@ function ilog2_2(i_in)
       ilog2_2=0
       i=i_in
       if(i<=0) return
-      if(iand(i,i-1)/=0) then
+      if(iand(i,i-1_2)/=0) then
          !write(0,*) 'iand i-1'
          ilog2_2=1
```
davegill: Do we need to change this assignment from
```
ilog2_2=1
```
to 
```
ilog2_2=1_2
```
All of the other integers are declared that way.

davegill:
@mgduda @jimbresch 
I am OK with this change. It addresses the compiler's complaint, which is legitimate.
We are never going to overflow an integer(kind=1) or integer(kind=2) with an assignment of "1", so my comments can be viewed that way.

ungrib/src/ngl/g2/intmath.f
mgduda: There should be no problem here to let the compiler perform the type conversion.

ungrib/src/ngl/g2/intmath.f
mgduda: There should be no problem here to let the compiler perform the type conversion.

davegill:
@mgduda 
Michael,
I am good with these changes

hnyc2020:
@davegill @mgduda 

I modified the intmath.f per your note as I could not execute ./ungrib.exe initially. Following your note, it worked!

So, the issue that I have with intmath.f file is that when I want to compile Nceplibs for UPP, I get the error as 


207 | if(iand(i,i-1)/=0) then
| 1
Error: Arguments of ‘iand’ have different kind type parameters at (1)
intmath.f:172:14:

172 | if(iand(i,i-1)/=0) then
| 1
Error: Arguments of ‘iand’ have different kind type parameters at (1)
make[1]: *** [makefile:108: ../../../../libg2_v3.1.0_4.a(intmath.o)] Error 1

Lines 207 and 172 are related to these updated lines. 

I have posted my question in more details here:
https://forum.mmm.ucar.edu/phpBB3/viewtopic.php?f=98&t=9818

I have attached my intmath.f file to this post.

I use Gfortran 9.3.0
WPS 4.1, WRF 4.1.5, Ubuntu 20 and NetCDF 4.1.2, Jasper 1.900.1, mpich 3.3.2, libpng 1.6.37, Zlib 1.2.11, 

I appreciate your thoughts on that.

Thank you!
[intmath.txt](https://github.com/wrf-model/WPS/files/5647779/intmath.txt)

mahseq:
TANKSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

mahseq:
hnyc2020 YOUR COMMAND IS VERY GOOOOOOOOOOOOD 
I CAN MAKE UNGRIB.EXE AFTER ONE WEEK
OH MY GOD 
THANKS A LOT

jose-rc84:
I really appreciate your help.

Thank you so much.

PR #121, opened 2019-03-30
Fix string comparison

kwadrat:
There is an error when == operator is used with single square brackets:

+ [ ../WRF == none ]
./configure: 183: [: ../WRF: unexpected operator

According to:
https://unix.stackexchange.com/questions/382003/what-are-the-differences-between-and-in-conditional-expressions

The = operator is the POSIX compatible way of comparing strings in sh/test.

I think there are 2 ways of comparing strings:
if [ $wrf_dir = "none" ]; then
if [[ $wrf_dir == "none" ]]; then

### DIFF
--- a/configure
+++ b/configure
@@ -180,7 +180,7 @@ else
    wrf_dir=$WRF_DIR
 fi
 
-if [ $wrf_dir == "none" ]; then
+if [ $wrf_dir = "none" ]; then
    echo ""
    echo "Error: No compiled WRF code found. Please check that the WRF model has been compiled"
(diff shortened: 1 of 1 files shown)


mgduda:
@kwadrat Agreed.

The development in the WPS repository takes place on branches, which are merged onto the `master` branch only when making a tagged release of the code. Accordingly, would you be willing to create a branch in your fork of the repository (e.g., named `fix_wrf_dir_comparison` or similar) with this change, and to create a new PR targeting the `release-v4.1` branch?

PR #122, opened 2019-04-02
Fix string comparison

kwadrat:
There is an error when == operator is used with single square brackets:

+ [ ../WRF == none ]
./configure: 183: [: ../WRF: unexpected operator

According to:
https://unix.stackexchange.com/questions/382003/what-are-the-differences-between-and-in-conditional-expressions

The = operator is the POSIX compatible way of comparing strings in sh/test.

I think there are 2 ways of comparing strings:
if [ $wrf_dir = "none" ]; then
if [[ $wrf_dir == "none" ]]; then

### DIFF
--- a/configure
+++ b/configure
@@ -180,7 +180,7 @@ else
    wrf_dir=$WRF_DIR
 fi
 
-if [ $wrf_dir == "none" ]; then
+if [ $wrf_dir = "none" ]; then
    echo ""
    echo "Error: No compiled WRF code found. Please check that the WRF model has been compiled"
    echo "       one directory level up (i.e., ../) in a directory named 'WRF', 'WRF-4.0.3', 'WRF-4.0.2', etc.,"


davegill:
@kwadrat 
Piotr,
Here is the syntax in the WPS/configure file:

```
if [ $wrf_dir == "none" ]; then
   echo ""
   echo "Error: No compiled WRF code found. Please check that the WRF model has been compiled"
   echo "       one directory level up (i.e., ../) in a directory named 'WRF', 'WRF-4.0.3', 'WRF-4.0.2', etc.,"
   echo "       or specify the full path to the compiled WRF model with the environment variable \$WRF_DIR ."
   echo ""
   exit
fi
```

Our systems have not complained about the combination of `sh`, `if`, `[`, and `==`. Thanks for pointing this out. Can you tell us what OS you are running?

kwadrat:
I tested it on:
1. Ubuntu 18.04.2 LTS 64-bit
2. Ubuntu 16.04.6 LTS
3. Ubuntu 12.04.5 LTS
where "/bin/sh" interpreter is linked to "dash"

There is no problem when "sh" is linked to "bash" but "dash" is the default shell on my Ubuntu systems.

I tested it also on very old "SCO OpenServer(TM) Release 5", there is also the same problem with "==", but no problem with '=".

This is a miminized code to show the problem on these systems:
`#!/bin/sh
wrf_dir=../WRF
if [ $wrf_dir == "none" ]; then
   echo "Error:"
fi
`

davegill:
@kwadrat 
Piotr,
Would you like to put together a PR to WPS to address this? Target the release-v4.1 as the base branch.

kwadrat:
I am a bit confused. This (my current change) is a "PR" (Pull Request). It is about  WPS repository. This PR is against release-v4.1 as the base branch. What is lacking from my side?

davegill:
@kwadrat 

> I am a bit confused. This (my current change) is a "PR" (Pull Request). It is about WPS repository. This PR is against release-v4.1 as the base branch. What is lacking from my side?

Piotr,
Sorry, I was dealing with an "issue" with another user, and got confused.

davegill:
@kwadrat 
Piotr,
I added one of your comments into your PR message.

kwadrat:
OK. Thank you!

davegill:
@kwadrat 
Piotr,
I searched in the WRF code, in the top level configure script (also sh). I did not find the [[ construct used there.

kwadrat:
Yes, that is correct. I also was not seeing the [[ construct in WRF code. I know this construct from my other scripts.

I only indicated that you could use single square brackets with single "=" (the old, POSIX way) OR extended version of comparing strings (for example in "bash") with double square brackets and "==". The former way is very compatible. The latter way is less popular among shells.

My idea was to convert "==" to "=" and use single square brackets - that way WPS code will be compatible with any "sh"/POSIX system.

mgduda:
@kwadrat Thanks very much for spotting this issue and for your thorough testing on different systems!

kwadrat:
Thank you very much!

PR #123, opened 2019-04-07
Update g2 lib for new HRRR fields, for g2print capability

jimbresch:
These mods to the g2 library's internal parameter table are specifically to identify variables for the
g2print capability. These modifications are not related to an upgrade for the ungrib Vtables.

Variable with a changed name:
CLMR

New variables:
HAIL
CIMIXR
FRICV
USWRF
MNUPHL
CNWAT
GFLUX
SSRUN
BGRUN

### DIFF
--- a/ungrib/src/ngl/g2/params.f
+++ b/ungrib/src/ngl/g2/params.f
@@ -40,7 +40,7 @@
 !
 !$$$
 
-      integer,parameter :: MAXPARAM=791
+      integer,parameter :: MAXPARAM=801
 
       type gribparam
           integer :: g1tblver
@@ -185,7 +185,7 @@
 !  Added 8/07/2003
 !
       data paramlist(128) /gribparam(2,229,0,0,192,'SNOHF')/
-      data paramlist(129) /gribparam(2,153,0,1,22,'CLWMR')/
+      data paramlist(129) /gribparam(2,153,0,1,22,'CLMR')/
       data paramlist(130) /gribparam(2,140,0,1,192,'CRAIN')/
       data paramlist(131) /gribparam(2,141,0,1,193,'CFRZR')/
       data paramlist(132) /gribparam(2,142,0,1,194,'CICEP')/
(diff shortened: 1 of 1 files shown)


davegill:
@mgduda @jimbresch 
Folks,
I approved this and then I was concerned ...
If we add these new variables into the ungrib stream, should we have an accompanying set of mods for each of these variables in METGRID.TBL? 

Here are the _new_ fields:
Changed name: 
CLMR

New:
HAIL
CIMIXR
FRICV
USWRF
MNUPHL
CNWAT
GFLUX
SSRUN
BGRUN

jimbresch:
Dave,

  These variables are for g2print. There are literally dozens of variables in the grib files that are not used by the WPS. If someone wants to use them, they would have to add them to the Vtable and METGRID.TBL themselves. 

Jim

davegill:
@jimbresch 
Jim,
Thanks for the clarification. Since these are for the g2print capability only, I am OK with a merge.

mgduda:
The changes seem fine, but I'd propose we update the description of this PR to explain that we're updating the g2 library internal tables used by g2print, and not Vtables.

mgduda:
Also, the commit message for cfbd28f has a couple of spelling errors that we could correct, and we might as well mention exactly which variable name has been amended.

mgduda:
In previous changes to the g2 library (e.g., 74eb62b1) we left in the old lines of code with comments. Should we do the same for these changes to the g2 library?

davegill:
@mgduda @jimbresch 
Michael,
When you mentioning leaving in the old values (commented out), this would be: old = CLWMR, and new = CLMR, correct? I can see an advantage for doing this with GFS data, as users absolutely use GFS for case studies. I do not know if HRRR data is also a resource that is used for IC/BC data for hindcasts.

If HRRR data is used retrospectively, I would be OK with commenting out the CLWMR line, and perhaps adding some approximate date info (of the naming transition) as well. If the maintenance of these "this variable is referenced this way between these dates" is anticipated to be overly onerous, I am OK with not including this documentation.

Last piece is ... can we push this until the v4.1.1 branch?

jimbresch:
There is no need to leave the previous line in the code. This change is to make WPS consistent with NCEP code. Only when WPS is changed to be different than the NCEP code do we need to document it.

PR #124, opened 2019-04-10
METGRID.TBL.ARW default pass QNC variable

gthompsnWRF:
There appears to be an errant output=no on the variable=QNC (cloud droplet number 
concentration) that can be removed from the metgrid/METGRID.TBL file. This is specifically
for those WPS users attempting to pass cloud mass and cloud number concentration from 
RAP/HRRR native model level files into real. 

The real program is able to correctly handle these fields due to the flags already in use in the
QNC stanza in the METGRID.TBL file.

### DIFF
--- a/metgrid/METGRID.TBL.ARW
+++ b/metgrid/METGRID.TBL.ARW
@@ -689,7 +689,6 @@ mpas_name=ni
         flag_in_output=FLAG_QNI
 ========================================
 name=QNC
-        output=no
         interp_option=four_pt+average_4pt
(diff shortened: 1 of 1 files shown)


davegill:
@mgduda 
Michael,
I don't see any troubles with this PR as it stands.
The QNC variable is not in any of the other METGRID.TBL files, but I do not think that we are responsible for maintaining those files.

mgduda:
After we change the target branch of this PR to `release-v4.1`, I think we'll be all set.

Issue #136, opened 2020-01-27
Fix for GHT bug in calc_ecmwf_p.exe 


hugohartmann:
We encountered corrupt GHT datafields provided by calc_ecmwf_p.exe. The problem occurs as we feed to calc_ecmwf_p only a subset of levels (24) we receive operationally from ECMWF. The corrupt GHT fields obviously result in erroneous WRF forecasts or even a crash of WRF.

Please find below the problem analysis:

The geopotential height (GHT) is calculated using the hypsometric equation (https://en.wikipedia.org/wiki/Hypsometric_equation). For that, the temperature (TT) and specific humidity (QV) are required.

Calc_ecmwf_p.exe computes the GHT at all 137 ECMWF hybrid levels, and needs TT and QV at all these levels. The program allocates memory for TT and QV 3D arrays and fills these arrays with input data.  Feeding a subset of the 137 levels into calc_ecmwf_p (we only have 24 levels), means that not all levels in the 3D array are filled. The program computes the TT and QV at the missing levels using linear interpolation. Missing levels are identified by looking at the TT-value at the first gridpoint (1,1) being equal to zero.

Now, memory allocation does not mean array initialization to zero! It can happen that the value at (1,1) is unequal to zero, meaning that the level is regarded as 'valid'. This leads to corrupt temperature and specific humidity data being transferred to the hypsometric equation, which results in corrupt GHT data.

Since the subsequent pre-processing utility mod_levs.exe removes all levels except for the subset of levels, one could argue that the GHT data at the subset of levels should be correct as the GHT data is only corrupt at the missing levels.

This is NOT the case. Calc_ecmwf_p.exe computes the GHT from level to level, starting at the lowest level 137. As a result, the computed GHT at the subset of levels is based on GHT on the missing levels with corrupt GHT values. 

Solution:
After memory allocation of TT and QV, force array initialization to zero.
tt(:,:,: ) = 0 & qv(:,:,: ) = 0 

In theory, this should suffice: tt(1,1,:) = 0

mgduda:
Thanks very much, @hugohartmann ! I've just merge PR #144 , which fixes this issue.

PR #144, opened 2020-04-21
Initialize arrays in calc_ecmwf_p to avoid potential error in GHT calculation

mgduda:
This PR fixes a potential error in the calculation of the GHT field in
the calc_ecmwf_p utility due to the use of uninitialized memory.

When the calc_ecmwf_p utility is provided with a subset of ECMWF model levels,
it may happen that the allocated 'tt' array contains non-zero values (because
it was not previously initialized). This leads to errors in the calculation of
the GHT field due to the use of uninitialized memory.

Thanks to @hugohartmann for identifying this issue and suggesting the fix in
this PR.

This PR fixes issue #136 .


hugohartmann:
You're welcome!

Issue #142, opened 2020-03-30
Ungrib not scaling negative starting latitudes


mgduda:
Typically, the starting latitude in a GRIB grid definition section is scaled by 1e6, and the ungrib code divides by this scaling factor. However, if the starting latitude is negative, no division by the scaling factor takes place in ungrib. For a discussion of this issue, see [this forum post](https://forum.mmm.ucar.edu/phpBB3/viewtopic.php?f=12&t=8978).

mgduda:
One fix that appears to work (see the forum post) is to change this code:
```fortran
              ! Scale lat/lon values to 0-180, default range is 1e6.
              if (map%lat1.ge.scale_factor) then
                 map%lat1 = map%lat1/scale_factor
              endif
```
around [line 494 of rd_grib2.F](https://github.com/wrf-model/WPS/blob/v4.1/ungrib/src/rd_grib2.F#L494-L497) to this:
```fortran
              ! Scale lat/lon values to 0-180, default range is 1e6.
              if (abs(map%lat1).ge.scale_factor) then
                 map%lat1 = map%lat1/scale_factor
              endif
```

mgduda:
@jimbresch Does the fix proposed here seem reasonable to you? If so, I'll create a PR to be included in the v4.2 release on Thursday.

davegill:
@mgduda 
Michael,
The `abs()` mod will work for GRIB data with < 1 million latitude points (plenty of room for data sets to get much larger). This seems like a simple fix that should not impact the existing positive lat1 values.

mgduda:
@davegill Why would this work only for datasets with < 1 M latitude points?

davegill:
@mgduda 
Here's what I was thinking: If the scale factor is about 1e3, and we have 1e6 grid points between the equator and the pole, and the lat1 point is just north of the equator, lat1 * scale_factor < scale_factor.

However, if these corner grid values are scaled so that they are always integers (which is probably what happens in the GRIB code), I guess that the scale factor will just get larger with the resolution.

mgduda:
The merge of PR #145 resolves this issue.

PR #145, opened 2020-04-21
Fix scaling of starting latitude in GRIB2 datasets on a Gaussian grid

mgduda:
This PR fixes an issue in scaling the starting latitude in GRIB2 datasets on 
a Gaussian grid.

Prior to the changes in this PR, the code in rd_grib2.F compared the starting
latitude for a field on a Gaussian grid with the scaling factor, and if 
the starting latitude was larger than the scaling factor, the starting latitude
was scaled. However, this fails when the starting latitude is negative, in which
case no scaling is applied, and downstream code fails due to an invalid starting
latitude in the resulting intermediate file.

The fix in this PR is to simply compare the magnitude of the starting latitude 
with the scaling factor, and if the former exceeds the latter, then scaling of 
the starting latitude is performed.

Thanks to Linda Maoyi for identifying the issue addressed in this PR.

This PR fixes issue #142 .

### DIFF
--- a/ungrib/src/g2print.F
+++ b/ungrib/src/g2print.F
@@ -562,7 +562,7 @@ end program g2print
               endif
 
               ! Scale lat/lon values to 0-180, default range is 1e6.
-              if (map%lat1.ge.scale_factor) then 
+              if (abs(map%lat1).ge.scale_factor) then 
                  map%lat1 = map%lat1/scale_factor
               endif
               if (map%lon1.ge.scale_factor) then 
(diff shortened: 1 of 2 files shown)


davegill:
@jimbresch 
Jim, I toyed around with this PR for a bit. The simple fix seems solid.

mgduda:
@jimbresch @davegill I've pushed an additional commit to this PR to handle scaling of negative starting latitudes in the `g2print` utility. Could one of you leave a GitHub review, which is needed before this PR can be merged?

Issue #132, opened 2019-07-17
geogrid generated landmask does not see lakes with modis_15s_lake


markelg:
There seems to be a "21" missing in GEOGRID.TBL in the landmask definition for modis_15s_lake

```
        landmask_water = modis_15s_lake:17         # Calculate a landmask from this field
        landmask_water =      modis_30s:17         # Calculate a landmask from this field
        landmask_water = modis_30s_lake:17,21      # Calculate a landmask from this field
```

This causes the LANDMASK computed by geogrid to not to see the lakes. I am not sure about the consequences, I think that real.exe fixes this afterwards recomputing the landmask, but at least this screws up my domain plotting script which uses geo_em files.

mgduda:
I agree that we seem to be missing category 21 from the `landmask_water` specification for `modis_15s_lake`.

mgduda:
@markelg Thanks very much for pointing out this issue. I've just merged PR #146 , which fixes this issue.

PR #146, opened 2020-04-21
Add missing category 21 for landmask calculation with modis_15s_lake land use

mgduda:
This PR adds category 21 as a water category for the modis_15s_lake land use to
give correct computation of the landmask field.

The 'landmask_water' entry in the GEOGRID.TBL.ARW file for the 'modis_15s_lake'
land use dataset previously only identified category 17 as a water category,
leading to an incorrect landmask field in the geogrid output.

This PR updates the 'landmask_water' entry to identify both category 17
and category 21 as water categories.

Thanks to @markelg for finding the issue addressed by this PR.

This PR fixes issue #132 .


davegill:
@weiwangncar 
Wei,
The lake category for the 15s resolution is the same as for the 30s resolution. This seems OK to me.

weiwangncar:
@mgduda So the lake category is in the dataset, but it just needs to be identified using GEOGRID.TBL?

And thank you for addressing all these WPS issues.

weiwangncar:
@mgduda OK, I see that we do have a 15 s dataset that has lakes.. I'm ok with this PR.

Issue #129, opened 2019-05-06
POWER/Linux configuration needs -qufmt=be


mgduda:
For compatibility with other systems, the binary files (principally, the intermediate files) created by the WPS program should be written in big-endian byte order. Accordingly, the `-qufmt=be` option should be added to the FFLAGS variable in the POWER/Linux stanza.

davegill:
@mgduda 
Michael,
Supporting your position:

On page 25 of http://geco.mines.edu/prototype/How_do_you_build_applications/xl/ppc/xlf/getstart.pdf is this note:

<img width="506" alt="Screen Shot 2019-05-06 at 5 21 25 PM" src="https://user-images.githubusercontent.com/12666234/57261059-8e97f380-7023-11e9-8cd8-ae2901c3435a.png">

mgduda:
This issue has been resolved as of the merge of PR #147 .

PR #147, opened 2020-04-22
Add -qufmt=be to FFLAGS and F77FLAGS for XL/POWER/Linux stanza

mgduda:
This PR adds the `-qufmt=be` option to FFLAGS and F77FLAGS in the XL/POWER/Linux
configuration stanza.

For compatibility with other systems, the binary files (principally,
the intermediate files) created by the WPS programs should be written in
big-endian byte order. Adding the -qufmt=be compiler option to the FFLAGS and
F77FLAGS variables in the POWER/Linux stanza accomplishes this.

This PR fixes issue #129 .

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -11,8 +11,8 @@ SFC                 = xlf2003_r
 CC                  = mpicc
 SCC                 = xlc_r
 LD                  = $(FC)
-FFLAGS              = -qfree=f90
-F77FLAGS            = -qfixed
+FFLAGS              = -qfree=f90 -qufmt=be
+F77FLAGS            = -qfixed -qufmt=be
 FNGFLAGS            = $(FFLAGS)
 LDFLAGS             =
 CFLAGS              =


mgduda:
Before the changes in this PR, creating an intermediate file on Summit and trying to use that intermediate file on my Linux/x86 desktop results in the following error:
```
ERROR: Didn't recognize format version of data in FILE_LE:2020-04-22_00.\nFound version 83886080 but expected either 3, 4, or 5. This could be an endian problem.
```
After the changes in this PR, I'm able to use an intermediate file created on Summit on my desktop system.

davegill:
@mgduda 
Michael,
Do you recommend that we do this for WRF also? There are binary files that are read, such as for the various radiation look-up tables.

I see this example xlf compiler stanza for WRF:
```
###########################################################
#ARCH    Linux x86_64 ppc64le i486 i586 i686, xlf compiler with xlc # serial smpar dmpar dm+sm
#
```
blah blah blah 
```
FCBASEOPTS_NO_G =       -w -qspill=81920 -qmaxmem=-1 $(FORMAT_FREE) $(BYTESWAPIO) 
```

Should we add the `-qufmt=be` flags?

mgduda:
@davegill We probably should do something like this for WRF as well. After some effort, I still wasn't able to get the entire WRF model to build on Summit with the XL compilers; I was able to get the I/O libraries to build, which was enough to allow me to build the WPS, however.

Would it be worth opening up an issue in the WRF repository to discuss and track WRF build issues on Summit?

davegill:
@mgduda 
Michael,
I'll follow up from a WRF repo

PR #155, opened 2020-06-07
[FIX] metgrid memory bug

### LINKED ISSUE #20: Double Precision WRF Build Causes Metgrid Failure
When WRF is built with double-precision, it causes metgrid to fail with error:
"Require grid spacing (dx) in meters to be positive!  ERROR:  MAP_INIT"
This error is located in module_map_utils.F
namelist.wps declares dx = 30000
print statements in module_map_utils.F confirms that dx is being read as 0.0000000E+00
map_proj = 'lambert'

lpilz:
This fixes #20, a memory bug caused by IO subroutines expecting arrays but receiving reals. Since this bug only occurred in double precision, this fix should be enough. It should not be necessary to change the `ext_get_dom_ti_integer_scalar` subroutines as well.

Thanks to @pblossey for pointing this out.


davegill:
@lpilz @mgduda
1. The title should be sufficient to give an indicator of what change was made, so "memory bug" is not sufficient. Perhaps mention double precision build, input of metadata, etc. Was the error related to the kind of real or related to a single value vs an array.
2. Would you please add a more thorough commit message. Several more sentences would be OK. From the commit message alone we want to have a fairly clear idea of what changes you made and why those changes were made. You can mention the original problem (dx). When you run ncdump on metgrid data (before the change vs after the change), do you see the fix in the metadata dx value? If so, those demonstration lines could be added as well.
3. These changes are to the input routines (the routine names all have "get" in them: for example, ext_int_get_dom_ti_real). Are any changes required for the output routines (the routines with "put" in the name). For example, are your mods sufficient to allow the real program to use the data from double precision metgrid?

davegill:
@mgduda 
Michael,
Do you want this PR to develop or maybe 4.2._x_

lpilz:
@davegill

I implemented your remarks 1&2. Please check if this is to your satisfaction. For 2, I referenced the issue - hope this is enough.
3. I'm not sure actually. You are right that the `ext_ncd_put_dom_ti_real` subroutine is similar, but I never encountered any issues with `real.exe` until now.

Whilst researching for your point 3. though, I found a typo, which might be related. In the call signature, the variable is named `OutCount`, while only `OutCOunt` is initialized. Cf. below

https://github.com/wrf-model/WRF/blob/0d88b5800186d892a882e118425d2e684ded756f/external/io_netcdf/wrf_io.F90#L1762

lpilz:
I did a bit of testing and found that for me, the metgrid output files don't differ between double and single precision. However, I would lean in the direction of including the `put` subroutines in this PR anyways, just to preempt any future problems. @davegill Is this okay?

mgduda:
@lpilz I think the issue addressed in this PR has been handled by PR #184 . I'll close this PR, but if it turns out that the changes as specified in this PR are necessary to fully address the issue, please feel free to re-open it.

PR #160, opened 2020-10-27
namelist.wps updates to newer case dates/domain to match updated default namelist.input

kkeene44:
Many settings for the default case in namelist.input were outdated or didn't follow current recommendations. A wrf-model PR ([#1306](https://github.com/wrf-model/WRF/pull/1306)) was submitted to update the namelist. To ensure consistency, updates are also made to namelist.wps to match those in namelist.input. Case dates, domain size, position, grid spacing, and latitude/longitude are modified. A note at the top of namelist.wps was added to inform users of the option to add comments. The new case (2019 Hurricane Dorian) was tested to ensure the outcome looks reasonable.

### DIFF
--- a/namelist.wps
+++ b/namelist.wps
@@ -1,8 +1,8 @@
 &share
  wrf_core = 'ARW',
  max_dom = 2,
- start_date = '2006-08-16_12:00:00','2006-08-16_12:00:00',
- end_date   = '2006-08-16_18:00:00','2006-08-16_12:00:00',
+ start_date = '2019-09-04_12:00:00','2019-09-04_12:00:00',
+ end_date   = '2019-09-06_00:00:00','2019-09-04_12:00:00',
  interval_seconds = 21600
  io_form_geogrid = 2,
 /
@@ -10,10 +10,10 @@
 &geogrid
  parent_id         =   1,   1,
  parent_grid_ratio =   1,   3,
- i_parent_start    =   1,  31,
- j_parent_start    =   1,  17,
- e_we              =  74, 112,
- e_sn              =  61,  97,
+ i_parent_start    =   1,  53,
+ j_parent_start    =   1,  25,
+ e_we              =  150, 220,
+ e_sn              =  130, 214,
  !
  !!!!!!!!!!!!!!!!!!!!!!!!!!!! IMPORTANT NOTE !!!!!!!!!!!!!!!!!!!!!!!!!!!!
  ! The default datasets used to produce the MAXSNOALB and ALBEDO12M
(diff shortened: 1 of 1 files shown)


namelist.wps
```diff
@@ -1,19 +1,21 @@
+! Add comments by prefacing with an "!" 
+
 &share
  wrf_core = 'ARW',
  max_dom = 2,
- start_date = '2006-08-16_12:00:00','2006-08-16_12:00:00',
- end_date   = '2006-08-16_18:00:00','2006-08-16_12:00:00',
+ start_date = '2019-09-04_12:00:00','2019-09-04_12:00:00',
+ end_date   = '2019-09-06_00:00:00','2019-09-06_00:00:00',
```
mgduda, line 7: Should we set `end_date` to the same as `start_date` for domain 2?

kkeene44: @mgduda 
Probably. That's the way we've done it in the past, but let's get @weiwangncar opinion.

smileMchen: Michael and Kelly,
I think we can set end_date to the same as start_date for D02. This is because we only needs met_em file at the initial time for D02.

weiwangncar: I am ok with that. Thanks.

namelist.wps
```diff
@@ -28,11 +30,11 @@
...
+ dx = 15000,
+ dy = 15000,
  map_proj = 'lambert',
- ref_lat   =  34.83,
- ref_lon   = -81.03,
+ ref_lat   =  33.00,
+ ref_lon   = -79.00,
  truelat1  =  30.0,
  truelat2  =  60.0,
  stand_lon = -98.0,
```
weiwangncar, line 40: @kkeene44 Should we change stand_lon to be the same as ref_lon? Can you add the new domain plot?

weiwangncar:
This is the domain map according to the proposed namelists:
wps_show_dom

namelist.wps
kkeene44: @weiwangncar Not sure how I missed this message from you. Yes, I agree. I made the change and committed it.

mgduda:
@kkeene44 Can you squash the four commits in this PR into one? You can do this with an interactive rebase; just let me know if I can help out with this.

kkeene44:
@mgduda I think I've done this correctly. Let me know if not!

mgduda:
@kkeene44 All looks good -- thanks!

mgduda:
@kkeene44 Please let me know when you're done modifying this PR, and I'll give it a final review before merging it.

kkeene44:
@mgduda I think I'm finally done. Thanks!

mgduda:
Just so that we have a record of what the new domains look like, here's an updated plot:
dorian_domains

If @weiwangncar and @smileMchen approve, then I'll merge this PR.

weiwangncar (approved):

mgduda (approved):

PR #165, opened 2021-01-29
Fix URL for getting NARR static grib1 data

davegill:
The information for users in the NARR Vtable accidentally includes
an additional letter after "www". Once this change is made, sure
enough, we can download the NARR static data.

This is not a compiled file, so this change has no impact on the source build.

This part of the NARR Vtable is not processed by ungrib, it is purely a comment.

Changes to be committed:
    modified:   ungrib/Variable_Tables/Vtable.NARR


davegill:
@jimbresch 
Jim,
You are the gate keeper for the ungrib portions of the code. This one is uncontroversial.

weiwangncar (approved):

mgduda (approved):

PR #169, opened 2021-04-20
Add 20 new GWD fields in this file

weiwangncar:
The new gravity wave drag option to be released with WRF v4.3 (gwd_opt = 3) requires 20 new input fields for VAR, CON, OA1-4, OL1-4. These are added in the new file GEOGRID.TBL.ARW_gwd-opt3. To include these fields in geogrid processing, append this file to GEOGRID.TBL.ARW or GEOGRID.TBL.

The first group of data is similar to those required for gwd_opt = 1, but with slight different processing (i.e. removing de-trending applied). The second group of data is new and needed to account for small-scale GWD and turbulent orographic form drag. 

The data is contributed by Mike Toy of NOAA/GSL.


weiwangncar:
We will make a different PR to add these new fields.

dudhia:
What is the original resolution of these GWD fields?

PR #175, opened 2021-05-05
Add --nowrf option to configure script, for configuring without WRF

mgduda:
This PR adds a new option, `--nowrf`, to the `configure` script so that the WPS
can be configured without a compiled WRF model.

Only three of the WPS components -- geogrid, metgrid, and int2nc -- require
the WRF I/O libraries. Even if none of these components was needed, it was
previously still necessary to either compile WRF or to make manual changes to
the `configure` script.

This PR introduces logic in the `configure` script to check for a command-
line argument, `--nowrf`, that, if present, disables checks in the `configure`
script for a compiled WRF directory. This enables, for example, the compilation
of the ungrib program without needing to compile the WRF model.

### DIFF
--- a/configure
+++ b/configure
@@ -1,6 +1,17 @@
 #!/bin/sh
 
-arg1=$1
+#
+# Check for command-line arguments
+# At present, the only supported argument is --nowrf, which disables checks to
+# find the path to the compiled WRF model.
+#
+nowrf=0
+for arg in $@; do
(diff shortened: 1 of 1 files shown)


weiwangncar:
@mgduda I tested it on Cheyenne with and without --nowrf, and it works as it is supposed to. I will approve this PR.

weiwangncar (approved):

PR #178, opened 2021-05-10
Updated namelist.wps for consistency with updates to namelist.input

kkeene44:
Updated namelist.wps for consistency with updates to namelist.input (https://github.com/wrf-model/WRF/pull/1513), as well as minor improvements to decrease user confusion.

- Set interval_seconds to 10800 - namelists are set up specifically for GFS data, which most types are available every three hours.
- Removed io_form_geogrid and io_form_metgrid variables, as currently we only recommend setting these to 2 (default)
- Removed comment regarding using V3.9 settings for geog_data_res (outdated)

### DIFF
--- a/namelist.wps
+++ b/namelist.wps
@@ -3,8 +3,7 @@
  max_dom = 2,
  start_date = '2019-09-04_12:00:00','2019-09-04_12:00:00',
  end_date   = '2019-09-06_00:00:00','2019-09-04_12:00:00',
- interval_seconds = 21600
- io_form_geogrid = 2,
+ interval_seconds = 10800
 /
 
 &geogrid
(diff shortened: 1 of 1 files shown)


weiwangncar:
The changes look ok to me. Thanks.

mgduda (approved):

davegill:
Just another set of eyes on this. The mods to WPS seem what we agreed to, and associated values for WRF are consistent.

kkeene44:
@andrewsoong Thank you for noticing the date discrepancy. It has now been corrected to 2021. As for WPS, it is being worked on as we speak. It's nearly impossible to release both at the exact same time since they are 2 different repositories and are released by different staff members. We expect it to be released by later this evening.

PR #181, opened 2021-09-21
Distribute and build zlib, png, and Jasper for ungrib

MiCurry:
This merge adds zlib, png, and Jasper libraries to be distributed and built with ungrib, rather than having the user needing to install all three of these libraries. zlib, png and Jasper all allow distribution under their similar open source licenses, as long as their licenses are included with their source code.

Before compiling ungrib, this merge currently compiles all of the above libraries in `ungrib/extern` and then builds ungrib by linking against those installations.

This PR serves as a draft PR. There has already been requested changes to this branch which include:

* Not compiling on other systems
    * Call make install on each library and linking against those libraries
* Use C compiler that was specified in the WPS `./configure`
* Specifying a flag to use distributed libraries, rather then compiling them by default.


weiwangncar:
This is good toward making the code more portable. Do we know if we need to worry about library updates?

MiCurry:
> This is good toward making the code more portable. Do we know if we need to worry about library updates?

These libraries are not update too often. These libraries are almost as old as Ungrib (and myself as well..) and my understanding of these libraries is that they are quite stable. The versions I added are not the newest, but I think we should be easily to update them to newer versions if we think we need a new release. Their build systems haven't changed since their release, so replacing one with a new version should be painless.

We may want to update zlib to version 1.6.37 for this PR, since it fixes a 'use-after-free' vulnerability. @mgduda do you think we should use the newest releases of PNG and JasPer? There is no reason that I picked the current libraries.

MiCurry:
Okay, this PR is ready to be reviewed by whoever would like to review it.

davegill:
@MiCurry 
The WPS geogrid and metgrid are built using the IO libraries from WRF. For a compressed NETCDF4 version of WRF, that would include zlib (which would be from the HPC default environment/settings/module). Are these new libs in the extern directory ONLY used for ungrib?

mgduda (changes requested):
I'm getting build errors on orcus with the `gnu/5.5.0` module:
```
CALLING ZLIB CONFIGURE AND MAKE
( cd zlib-1.2.7 ; CC= CPP=/usr/bin/cpp ./configure --prefix=/dev/shm/WPS/ungrib/extern && make MAKEFLAGS= && make install MAKEFLAGS= )
Checking for gcc...
Checking for shared library support...
No shared library support.
Building static library libz.a version 1.2.7 with .
Checking for off64_t... No.
Checking for fseeko... No.
Checking for strerror... No.
Checking for unistd.h... No.
Checking for stdarg.h... No.
Checking whether to use vs[n]printf() or s[n]printf()... using s[n]printf().
Checking for snprintf() in stdio.h... No.
  WARNING: snprintf() not found, falling back to sprintf(). zlib
  can build but will be open to possible buffer-overflow security
  vulnerabilities.
Checking for return value of sprintf()... No.
  WARNING: apparently sprintf() does not return a value. zlib
  can build but will be open to possible string-format security
  vulnerabilities.
Looking for a four-byte integer type... Not found.
make[2]: Entering directory `/dev/shm/WPS/ungrib/extern/zlib-1.2.7'
O -DNO_FSEEKO -DNO_STRERROR -DNO_snprintf -DHAS_sprintf_void -I. -c -o example.o test/example.c
make[2]: O: Command not found
make[2]: [example.o] Error 127 (ignored)
```
It looks like `CC` is getting set to an empty string here, which may not be what we want.

MiCurry:
> It looks like `CC` is getting set to an empty string here, which may not be what we want.

It looks like this was being caused by the ungrib/Makefile not passing $SCC into the extern directory (othermakefile variables get their $SCC passed by the `configure.wps` script). I just pushed a fix which should pass $SCC down to the `ungrib/extern` makefile.

As well, I have updated this branch to be based off of develop.

> @MiCurry The WPS geogrid and metgrid are built using the IO libraries from WRF. For a compressed NETCDF4 version of WRF, that would include zlib (which would be from the HPC default environment/settings/module). Are these new libs in the extern directory ONLY used for ungrib?

@davegill, yes currently, the extern directory is only used for ungrib, but you bring up a good point. Do most systems have zlib and pnglib installed by default? If so should we try and use them instead of compiling them?

mgduda (changes requested):
With the default modules on orcus
```
Currently Loaded Modules:
  1) ncarenv/1.0   2) ncarbinlibs/1.2   3) gnu/5.5.0   4) ncarcompilers/1.0   5) netcdf/4.7.4
```
I'm not able to build ungrib with the following
```
./configure --nowrf --build-extern
[choose option 1]
./compile ungrib
```
There seems to be some problem building the libraries, e.g., for zlib, I'm seeing the following
```
make[2]: Entering directory '/dev/shm/WPS/ungrib/extern/zlib-1.2.7'
gcc -O3  -D_LARGEFILE64_SOURCE=1 -DHAVE_HIDDEN -I. -c -o example.o test/example.c
ar rc libz.a adler32.o crc32.o deflate.o infback.o inffast.o inflate.o inftrees.o trees.o zutil.o compress.o uncompr.o gzclose.o gzlib.o gzread.o gzwrite.o
ar: adler32.o: No such file or directory
make[2]: [Makefile:128: libz.a] Error 1 (ignored)
```
and I'm not sure that the include search paths are correct
```
gcc -c  -D_UNDERSCORE -DBYTESWAP -DLINUX -DIO_NETCDF -DBIT32 -DNO_SIGNAL  -I /dev/shm/WPS/extern/include -DUSE_JPEG2000 -DUSE_PNG -D__64BIT__ dec_jpeg2000.c
dec_jpeg2000.c:5:27: fatal error: jasper/jasper.h: No such file or directory
compilation terminated.
make[2]: [Makefile:75: dec_jpeg2000.o] Error 1 (ignored)
```

MiCurry:
@mgduda  that's very interesting. I'll try and recreate your setup on Orcus with the same modules. It looks like make is being passed the `-r` option (ignore rules) in the zlib directory, which is why its trying to build the example. Just to confirm, you cloned your WPS into `/dev/shm`? And you're using the make at `/usr/bin/make`, which is version 3.82?

mgduda:
@MiCurry That's correct: my clone is in `/dev/shm/WPS`. I originally tried with `make` 3.82, but I've also tried using `make` 4.3 with no success.

Issue #180, opened 2021-06-28
Builds with GNU 11 compilers fail with argument type mismatch errors


mgduda:
When using the GNU 11 compilers, various errors in actual argument types prevent all three of the main WPS components (geogrid, ungrib, and metgrid) from compiling. For example:
```
gfortran  -ffree-form -O -fconvert=big-endian -frecord-marker=4 -c output_module.f90 -I/glade/scratch/duda/wrf_support/WPS/../WRF/external/io_netcdf -I/glade/scratch/duda/wrf_support/WPS/../WRF/external/io_grib_share -I/glade/scratch/duda/wrf_support/WPS/../WRF/external/io_grib1 -I/glade/scratch/duda/wrf_support/WPS/../WRF/external/io_int -I/glade/scratch/duda/wrf_support/WPS/../WRF/inc -I/glade/u/apps/ch/opt/netcdf/4.8.0/gnu/11.1.0//include
output_module.f90:1735:41:

 1735 |                                          var_value, &
      |                                         1
......
 1763 |                                          var_value, &
      |                                         2
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
output_module.f90:1680:41:

 1680 |                                          var_value, &
      |                                         1
......
 1708 |                                          var_value, &
      |                                         2
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
```

mgduda:
As a workaround, adding `-fallow-argument-mismatch` to the definition of `FFLAGS` and `F77FLAGS` allows the main WPS components to be compiled.

Issue #163, opened 2021-01-15
 Error compiling with gfortran 10


letmaik:
Same issue as in https://github.com/wrf-model/WRF/issues/1250.

```
2021-01-15T07:41:27.2436484Z gfortran  -ffree-form -O -fconvert=big-endian -frecord-marker=4 -c output_module.f90 -I/home/vsts/work/1/s/../WRF/external/io_netcdf -I/home/vsts/work/1/s/../WRF/external/io_grib_share -I/home/vsts/work/1/s/../WRF/external/io_grib1 -I/home/vsts/work/1/s/../WRF/external/io_int -I/home/vsts/work/1/s/../WRF/inc -I/home/vsts/work/1/s/netcdf/include
2021-01-15T07:41:27.2954031Z output_module.f90:1735:41:
2021-01-15T07:41:27.2955024Z 
2021-01-15T07:41:27.2955711Z  1735 |                                          var_value, &
2021-01-15T07:41:27.2956555Z       |                                         1
2021-01-15T07:41:27.2957208Z ......
2021-01-15T07:41:27.2957861Z  1763 |                                          var_value, &
2021-01-15T07:41:27.2958671Z       |                                         2
2021-01-15T07:41:27.2960372Z Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
```

anikfal:
@letmaik 
Could you work around this problem?

anikfal:
A temporary solution:
After configuring WPS, open `configure.wps` and add `-fallow-argument-mismatch` and `-fallow-invalid-boz` to the end of `FFLAGS` and `F77FLAGS` parameters. Now, after compiling, your will have all the exe files.

tnapoles:
I did the configure.wps fix above, geogrid.exe and ungrib.exe were created but metgrid.exe wasn't.
I see a bunch of errors for missing mod files which start with this:

Fatal Error: Cannot read module file ‘/localhome/wrf-build/netcdf/include/netcdf.mod’ opened at (1), because it was created by a different version of GNU Fortran. 

This is at the head of the netcdf.mod file 

GFORTRAN module version '10' created from netcdf.f90

I think I am using the correct version of gfortran, gcc 10 is the 1st gcc in my PATH though it's not in /usr/bin. Is it possible that WPS is using /usr/bin/gcc?

PR #184, opened 2021-11-02
Fix several incorrect argument ranks in calls to WRF I/O API routines

mgduda:
This PR corrects the actual argument rank in several calls to WRF I/O API
routines for reading and writing domain time-independent attributes.

### DIFF
--- a/geogrid/src/output_module.F
+++ b/geogrid/src/output_module.F
@@ -1429,21 +1429,21 @@ module output_module
 #ifdef IO_BINARY
       if (io_form_output == BINARY) then
          call ext_int_put_dom_ti_integer(handle, trim(var_name), &
-                                         var_value, &
+                                         (/ var_value /), &
                                          1, istatus)
       end if
 #endif
 #ifdef IO_NETCDF
       if (io_form_output == NETCDF) then
          call ext_ncd_put_dom_ti_integer(handle, trim(var_name), &
-                                         var_value, &
+                                         (/ var_value /), &
                                          1, istatus)
       end if
 #endif
 #ifdef IO_GRIB1
       if (io_form_output == GRIB1) then
          call ext_gr1_put_dom_ti_integer(handle, trim(var_name), &
-                                         var_value, &
+                                         (/ var_value /), &
                                          1, istatus)
       end if
 #endif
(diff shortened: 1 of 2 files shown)


mgduda:
This PR addresses issues pointed out in Issue #163 and in Issue #180.

davegill:
@mgduda 
Michael,

Is there a reason why some of these arrays you went the direction of `(/ var_value /)` and some are `var_value_arr`?

I ask, because likely what you do here we will need to conventionally put into the WRF model, as we have the exact same complaints of argument mismatches with the newer GNU compilers.

davegill:
@mgduda 
Michael,
OK, I got it now.
On "put", you force the input scalar to be an array.
On"get" you store the output variable into an array (which we later pull out index 1.

mgduda:
@davegill The details of the changes are explained in the commit messages:
https://github.com/wrf-model/WPS/pull/184/commits/03e35b60abbec9e06a2617313bcbdc52b14bb353
https://github.com/wrf-model/WPS/pull/184/commits/166c9886fb201865d2b3fd71c6cba22fbe1cde0d

I didn't think the PR description needed to go into such detail, though.

davegill:
@mgduda 
Michael,

I tested the following from inside the standard WRF-system docker container (GNU 9.3.1). This is expected to work since gnu 9 does not make mismatched variables a fatal error.
1. The standard WPS code builds (master branch).
2. This PR's branch (fix_ioapi_put_get_args) builds 

I tested the following with GNU 11.1.0 on cheyenne, where I expected to see an impact.
1. The standard WPS code does not build (master branch) for geogrid, ungrib, or metgrid.
2. This PR's branch (fix_ioapi_put_get_args) builds geogrid.exe and metgrid.exe (what was expected).

davegill (approved):

mgduda:
@davegill Once this PR has been merged, I was planning to create one more PR for the v4.3.1 release that adds `-fallow-argument-mismatch` to the GNU Fortran flags; that should allow everything to build, but I first wanted to address legitimate issues with subroutine arguments before papering over the remaining errors (e.g., calling `MPI_Bcast` with different argument types, which is completely fine and unavoidable).

davegill:
@mgduda 
Michael, 
I like your 2-PR idea.

PR #187, opened 2021-11-03
Fix builds with newer GNU compilers

mgduda:
This PR addresses build failures when newer GNU compilers are used.

Newer GNU compilers (in particular, major versions later than 9) appear to
require the -fallow-argument-mismatch option in order to successfully compile
certain modules that make calls to Fortran 77 routines that operate on various
datatypes: for example, the MPI_Bcast call can broadcast various data types, and
without the Fortran 90+ module interface, there are no overloaded explicit
interfaces for supported argument types, leading newer GNU Fortran compiler
releases to generate errors about inconsistent/mismatched argument types.

The configure script now builds a test program that determines whether the
Fortran compiler being used is GNU Fortran with a major version later than 9,
and sets the FCCOMPAT variable to -fallow-argument-mismatch in the configure.wps
file if so (and sets FCCOMPAT to an empty string otherwise).

Since Make treats undefined variables as empty strings, the FCCOMPAT variable
only needs to be defined for configuration options where it will be needed,
i.e., configuration options that use the GNU Fortran compiler.

Other compiler could follow what has been done in this commit to define FCCOMPAT
as needed through the use of additional test programs (and through the addition
of FCCOMPAT to other configuration stanzas).

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -159,6 +159,7 @@ CC                  = CONFIGURE_CC
 LD                  = $(FC)
 FFLAGS              = -ffree-form -O -fconvert=big-endian -frecord-marker=4
 F77FLAGS            = -ffixed-form -O -fconvert=big-endian -frecord-marker=4
+FCCOMPAT            = CONFIGURE_COMPAT_FLAGS
 FCSUFFIX            = 
 FNGFLAGS            = $(FFLAGS)
 LDFLAGS             =
@@ -182,6 +183,7 @@ CC                  = CONFIGURE_CC
 LD                  = $(FC)
 FFLAGS              = -ffree-form -O -fconvert=big-endian -frecord-marker=4
 F77FLAGS            = -ffixed-form -O -fconvert=big-endian -frecord-marker=4
+FCCOMPAT            = CONFIGURE_COMPAT_FLAGS
 FCSUFFIX            = 
 FNGFLAGS            = $(FFLAGS)
 LDFLAGS             =
@@ -452,6 +454,7 @@ CC                  = CONFIGURE_CC
 LD                  = $(FC)
 FFLAGS              = -ffree-form -O -fconvert=big-endian -frecord-marker=4
 F77FLAGS            = -ffixed-form -O -fconvert=big-endian -frecord-marker=4
+FCCOMPAT            = CONFIGURE_COMPAT_FLAGS
 FCSUFFIX            = 
 FNGFLAGS            = $(FFLAGS)
 LDFLAGS             =
@@ -477,6 +480,7 @@ CC                  = CONFIGURE_CC
 LD                  = $(FC)
 FFLAGS              = -ffree-form -O -fconvert=big-endian -frecord-marker=4
 F77FLAGS            = -ffixed-form -O -fconvert=big-endian -frecord-marker=4
+FCCOMPAT            = CONFIGURE_COMPAT_FLAGS
 FCSUFFIX            =
 FNGFLAGS            = $(FFLAGS)
 # For a WRF OpenMP build, add the gomp library for WPS
--- a/arch/postamble
+++ b/arch/postamble
@@ -18,10 +18,10 @@ AR		=	ar ru
 
 .f.o:
 	$(RM) $@ $*.mod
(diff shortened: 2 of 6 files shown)


mgduda:
@davegill Based on your experience with the WRF configuration script, if you have any suggestions for how to make the new test code in the WPS configure script more robust, I'd be glad to hear them.

mgduda:
@davegill I just found that the behavior of sed's `-i` option is slightly different on macOS, so I'm going to do what is done elsewhere in the configure script with
```
sed "s/CONFIGURE_COMPAT_FLAGS/${compat}/" configure.wps > configure.wps.tmp
mv configure.wps.tmp configure.wps
```
An updating branch is on its way.

davegill:
@mgduda 
Michael,
1. The FCCOMPAT Makefile macro is filled when using GNU versions 10 and 11 (standard exes are made)
2. GNU 7, 8, 9 does not have the FCCOMPAT macro filled in (standard exes are made for 9, didn't try others as WRF may not work with them)
3. I tried a couple Intel builds also - all fine. No FCCOMPAT line for the macro definition - again expected.

This seems solid

davegill (approved):

mgduda:
@davegill I actually have a variant that might be a bit more robust; here's the relevant snippet from the `configure` script:
```
# The above test program gives exit status 1 iff we are using the GNU Fortran
# compiler with a major version greater than 9

FC=`grep ^SFC configure.wps | cut -d"=" -f2-`
FFLAGS=`grep ^FFLAGS configure.wps | cut -d"=" -f2-`
$FC ${FFLAGS} gnu_flag_test.F90 -o gnu_flag_test > /dev/null 2>&1
if [ -f ./gnu_flag_test ]; then
   ./gnu_flag_test > /dev/null 2>&1
   if [ $? -eq 1 ]; then
      compat="-fallow-argument-mismatch"
   fi
   rm gnu_flag_test
else
   printf "*** Failed to compile the gnu_flag_test program!\n"
   printf "    This may be because the selected build option does not work correctly.\n"
fi
rm gnu_flag_test.F90

sed "s/CONFIGURE_COMPAT_FLAGS/${compat}/" configure.wps > configure.wps.tmp
mv configure.wps.tmp configure.wps
```
The key changes from what's currently in the PR are:
1. We re-obtain FC and FFLAGS in case earlier tests where these would ordinarily have been set are skipped (e.g., when the user chooses a non-functioning configure opiton)
2. We only try to run (and later remove) the `gnu_flag_test` program if it was actually compiled
3. If the `gnu_flag_test` program wasn't built, print an error message
4. Use integer comparison `-eq` rather than string comparison `=` when comparing `$?` with `1`

davegill:
@mgduda 
These updates seem even _more_ robust. Once that commit is in, I'll run through a couple of tests again. I know with WRF that once there are mods, the previously existing approvals are voided. This does not seem to be the case with WPS.

mgduda:
@davegill I've just pushed the updated configure script.

davegill:
@mgduda 
Michael,
I notice that the flag `$(FCCOMPAT)` is added only to the ungrib Makefiles. Does the top-level configure.wps build rules handle the others?

mgduda:
> @mgduda Michael, I notice that the flag `$(FCCOMPAT)` is added only to the ungrib Makefiles. Does the top-level configure.wps build rules handle the others?

Right -- the geogrid and metgrid Makefiles pick up the suffix rules from the configure.wps file with lines like
```
include ../../configure.wps
```

davegill:
@mgduda 
Michael,
I grabbed the most recent WRF release. For each WPS compile option listed below, I built the WRF model with the appropriate `configure` option, and with `compile io`. Then I went to the WPS directory and conducted a build test.
   . I tested these mods in a container (GNU 9.3.0), and all is well.
   . On cheyenne, I tested the newest Intel and Nvidia compilers (both from 2021). All builds well with WPS.
   . Also on cheyenne, I did just a quick retest of GNU9, 10, and 11. All build OK. `FCCOMPAT = -fallow-argument-mismatch` shows up when expected in configure.wps.

You seem to have pretty thoroughly tested this mod. 

This re-review is ACCEPT.

mgduda:
Thanks for the thorough testing, @davegill !

PR #193, opened 2022-01-26
Synchronize with WRF: no more NMM support

davegill:
Because the WRF model ended support for the NMM dynamical core with
release 4.3, the top-level and user-interface files in WPS have the NMM
information removed, beginning with the WPS release v4.4.

M README
D geogrid/GEOGRID.TBL.NMM
D namelist.wps.nmm


davegill:
@mgduda 
Michael,
I am thinking of something like the above for the WPS task of removing NMM support. If the namelist is removed, the geogrid TBL file is removed, and there is no mention of NMM in the top-level README - I think that is probably sufficient. This would be part of the v4.4 release in April.

mgduda (approved):
I've added a commit to remove the `METGRID.TBL.NMM` and `METGRID.TBL.NMM.rap` files, too.

PR #196, opened 2022-03-08
Fixing bug in calc_dfdx and calc_dfdy

Fergui:
TYPE: bug fix

KEYWORDS: subgrid, fire, terrain, slope, gradient

SOURCE: Timothy W. Juliano and Branko Kosovic (NCAR/RAL), Angel Farguell (San Jose State University)

DESCRIPTION OF CHANGES:
Problem:
When processing terrain data for the WRF-Fire subgrid (variable `ZSF`) in `geogrid`, the terrain gradients in the x- and y-directions (variables `dzdxf` and `dzdyf`) are computed from the `ZSF `field. At present, the `geogrid/src` code (`process_tile_module.F`) uses the meteorological grid cell spacing from the outermost domain, rather than the subgrid (fire mesh) cell spacing from the present domain, to compute the gradients. The resultant `dzdxf` and `dzdyf` fields are computed correctly if `subgrid_ratio_x=subgrid_ratio_y=1` and if `max_dom=1`; however, if `subgrid_ratio_x>1` on the fire mesh and/or `max_dom>1` and the fire mesh is in a child domain, then the resultant `dzdxf` field is computed incorrectly and is an underestimation of the actual terrain gradient. The issue is similar for the `dzdyf` field when `subgrid_ratio_y>1`. As the `subgrid_ratio` increases, the underestimation becomes more pronounced.

Solution:
Use the present domain meteorological grid cell spacing and account for the `subgrid_ratio` when computing the terrain gradients. This simply requires calling the `get_domain_resolution` subroutine to pull the present grid cell spacing of the meteorological domain, passing in the respective `subgrid_ratio` to the `calc_dfdx` and `calc_dfdy` subroutines, and dividing the correct meteorological grid spacing by `subgrid_ratio`.

LIST OF MODIFIED FILES:
M geogrid/src/process_tile_module.F

TESTS CONDUCTED:
Run `geogrid` before and after the code modifications and inspect variables `dzdxf` and `dzdyf` to ensure magnitudes are correct compared to the source SRTM data. [Note: We do not plot the comparisons on a georeferenced grid and so the SRTM data are shifted/rotated compared to the WRF outputs. This is meant to be a rough comparison.]

`dzdxf`:
all_dzdx

`dzdyf`:
all_dzdy

RELEASE NOTE:
When processing terrain data for the WRF-Fire subgrid, terrain gradient variables `dzdxf` and `dzdyf` are computed using the meteorological grid cell spacing from the outermost domain, thus resulting in an underestimation in gradient magnitude when `max_dom>1` or the `subgrid_ratio` for the fire mesh is greater than 1. Therefore, we correctly account for the current domain resolution and the `subgrid_ratio` when computing the terrain gradient variables.

### DIFF
--- a/geogrid/src/process_tile_module.F
+++ b/geogrid/src/process_tile_module.F
@@ -2046,7 +2046,7 @@ module process_tile_module
                         end_mem_i, end_mem_j, end_mem_k, sr_y, mapfac)
    
       ! Modules
-      use gridinfo_module
+      use llxy_module
     
       implicit none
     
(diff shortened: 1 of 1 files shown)


mgduda (approved):
This looks good to me.

@Fergui Thanks for the excellent PR description and figures!

PR #199, opened 2022-03-15
Update configure

MCDLnakamura:
This could not work for file name of  "WRF-4.3.3". I added latest version of WRF name.


weiwangncar:
@MCDLnakamura This should be placed under Issues. Not PR.

PR #206, opened 2022-05-19
Adding entries to the Metgrid table for MAD-WRF

pedro-jm:
MAD-WRF requires the following entries in the metgrid table. The variables are optional and can be used to improve the cloud analysis in Real.

### DIFF
--- a/metgrid/METGRID.TBL.ARW
+++ b/metgrid/METGRID.TBL.ARW
@@ -1148,3 +1148,19 @@ name=LOGSFP
         output=no
         interp_option=four_pt
 ========================================
+name=CLDMASK
+        interp_option=nearest_neighbor
+        flag_in_output=FLAG_CLDMASK
+========================================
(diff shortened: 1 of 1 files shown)


pedro-jm:
@mgduda @dudhia 
I am wondering if it would be possible to add this update to the Metgrid table sooner than WPS version 4.5 since that requires waiting pretty much a whole year. The change in the Metgrid table is unlikely to have side effect on WPS users, but not having the update has a major impact on MAD-WRF users.

weiwangncar:
@pedro-jm We can try 4.4.1.

pedro-jm:
@weiwangncar 
That would be good, thanks.

weiwangncar:
@mgduda Would you have time to make this part of 4.4.1 release?

mgduda (approved):
@pedro-jm Apologies for the long delay in getting to this PR!

PR #210, opened 2022-12-22
Adds GFS total column ozone field to metgrid table and creates new Vtable with GFS ozone field.

jordanschnell:
Adds GFS total column ozone field to metgrid table and creates new Vtable with GFS ozone field.

M       metgrid/METGRID.TBL.ARW
A       ungrib/Variable_Tables/Vtable.GFS_OZONE

### DIFF
--- a/metgrid/METGRID.TBL.ARW
+++ b/metgrid/METGRID.TBL.ARW
@@ -761,8 +761,6 @@ name=T0
         missing_value=200.
         fill_missing=-20.
 ========================================
-name=O3_TOT ; output_name=O3_GFS_DU
-========================================
 name=O3_GFS_DU
         interp_option=four_pt+average_4pt
         fill_lev=200100:const(200100.)
--- a/metgrid/METGRID.TBL.ARW
+++ b/metgrid/METGRID.TBL.ARW
@@ -761,6 +761,13 @@ name=T0
         missing_value=200.
         fill_missing=-20.
 ========================================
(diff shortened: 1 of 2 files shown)


mgduda:
@jordanschnell Curiously, using `O3_GFS_DU` as the variable name in the `Vtable` doesn't work (as you pointed out offline). I'll take a quick look into the ungrib code to see if I can figure out why this is happening. If there is an easy fix, perhaps we could modify this PR; but if a fix to ungrib seems non-trivial, I think moving ahead with this PR as-is should be fine.

mgduda:
I think the issue with nine-character variable names is becoming clearer (see Issue #215), and a fix seems not too difficult. I'll put together a PR to address this, and after that's been merged, we could update this PR to avoid the need for an "intermediate" variable name of `O3_TOT`.

mgduda:
@jordanschnell With the merge of PR #216, I think it should now be possible to remove the
```
========================================
name=O3_TOT ; output_name=O3_GFS_DU
========================================
```
entry from the `METGRID.TBL.ARW` file and to use `O3_GFS_DU` in the new `Vtable.GFS_OZONE`. If you see any issues, though, I'll be glad to investigate.

jordanschnell:
@mgduda Sounds good - The cluster I work with the code on is down today, but I'll get it updated as soon as I can.

mgduda (approved):

PR #212, opened 2022-12-23
Add -h/--help options to the 'configure' script

mgduda:
This PR adds `-h`/`--help` options to the configure script to print a summary of the available
configure options. The message reads as follows:
```
Usage: configure [options]

where options may be any of the following:

 --nowrf             Disables checks to find the path to the compiled WRF model.
                     This enables some WPS programs to be compiled without
                     requiring WRF to have been compiled, but it will preclude
                     the compilation of WPS programs that depend on the WRF I/O
                     API (geogrid, metgrid, and int2nc).

 --build-grib2-libs  Compiles zlib, libpng, and JasPer libraries from source in
                     the external/ directory, and installs the libraries in
                     grib2/.

 -h/--help           Print this help message and quit.
```

### DIFF
--- a/configure
+++ b/configure
@@ -1,5 +1,24 @@
 #!/bin/sh
 
+print_help()
+{
+  echo 'Usage: configure [options]'
+  echo ''
+  echo 'where options may be any of the following:'
+  echo ''
+  echo ' --nowrf             Disables checks to find the path to the compiled WRF model.'
+  echo '                     This enables some WPS programs to be compiled without'
+  echo '                     requiring WRF to have been compiled, but it will preclude'
+  echo '                     the compilation of WPS programs that depend on the WRF I/O'
+  echo '                     API (geogrid, metgrid, and int2nc).'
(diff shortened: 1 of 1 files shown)


kkeene44:
@mgduda 
I think it looks good! I wasn't aware of the --nowrf option, so that's good to know. Would it be beneficial to add the option to include -d or -D for debugging? There haven't been many times debugging has been necessary for WPS, though, so maybe not.

smileMchen (approved):

kkeene44 (approved):

mgduda:
> @mgduda I think it looks good! I wasn't aware of the --nowrf option, so that's good to know. Would it be beneficial to add the option to include -d or -D for debugging? There haven't been many times debugging has been necessary for WPS, though, so maybe not.

@kkeene44 We could certainly implement the capability to create debug builds of the WPS in a future PR, but for this PR, the focus is on documenting existing options with a help message.

PR #214, opened 2023-01-05
add global LCZ entry to GEOGRID.TBL

cenlinhe:
This update is to add the global 100-m LCZ dataset entry to GEOGRID.TBL by creating a new GEOGRID.TBL.ARW_LCZ. When users would like to use the LCZ capability, they need to use the new GEOGRID.TBL.ARW_LCZ table.


mgduda:
@cenlinhe Just one idea to consider: I think if you add `optional=yes` to the `priority=2` entry for `LANDUSEF`, it would be possible to add the new entry to the existing `GEOGRID.TBL.ARW` file -- i.e., we wouldn't necessarily need a new `GEOGRID.TBLARW_LCZ` file. The downside is that this generates a message at runtime that might confuse users:
```
Parsed 51 entries in GEOGRID.TBL
Processing domain 1 of 1
  Processing XLAT and XLONG
  Processing MAPFAC
  Processing F and E
  Processing ROTANG
  Processing LANDUSEF
   Important note: could not open input dataset for priority level 2, but this level is optional.
  Calculating landmask from LANDUSEF ( WATER =  17  21 )
  Processing HGT_M
  ...
```
Perhaps in a future release we could move to a unified `GEOGRID.TBL.ARW` after we make the "Important note" message a bit clearer in geogrid.

dudhia:
The message may be OK if Important was removed from it.

cenlinhe:
@mgduda Do you prefer me to modify this PR to have the unified GEOGRID.TBL.ARW or keep it as it is, then creating a unified GEOGRID.TBL.ARW in the future?

mgduda:
@cenlinhe I think for now we could keep this PR as is. For a future release (or even the v4.5 release if there's time; recall that we can create the WPS `release-v4.5` branch later than WRF creates its release branch) we can discuss options for supporting a unified `GEOGRID.TBL.ARW` file.

cenlinhe:
sounds good to me!

mgduda (approved):

PR #154, opened 2020-05-27
Adding support for ICON-EU in WPS

grafmi:
The ICON-EU grib2 data set is now able to be ingested by the ungrib program.
1. New Vtable:
   - A new Vtable file for ICON-EU (EU-Nest) has been created with the fields required by the
WRF model.
2. Information about the data:
   - The input data (grib2) of ICON-EU can be 
found here: https://opendata.dwd.de/weather/nwp/icon-eu/grib/ 
   - You will find the documentation about ICON and ICON-EU here: 
https://www.dwd.de/DWD/forschung/nwv/fepub/icon_database_main.pdf 
3. Modifications to the ungrib source code:
   - I added support for level code 150 in rd_grid2.F because model levels of ICON-EU are 
on generalized vertical height coordinates. The general vertical height coordinate is not 
specified by vertical coordinate parameters but by providing a 3D field, that specifies the 
geometric height of every model grid point in meters. For more details, see: 
http://www.cosmo-model.org/content/model/modules/coding/grib/gribVerticalCoordinates.pdf
4. Addition of new soil levels:
   - New soil levels from ICON-EU are added to METGRID.TBL.ARW for metgrid.exe. Details 
about the soil levels in ICON-EU can be found in the documentation (see above)
5. Conversion of soil moisture units to those expected by the WRF model:
   - The preprocessing in rrpr.F has to be modified to convert the units of column-integrated 
soil moisture provided by ICON-EU from kg m-2 to m3 m-2 for WRF input. 
## At the moment, I'm using a layer which exists only in ICON-EU as trigger to do the conversion. Maybe, there's a better solution. 

I have performed two model runs with exactly the same configuration and input from GFS 
and ICON-EU (init time 20200531_00 and lead time 72 h). The output of both runs are similar, 
but it can be recognized, e. g., that there is less shallow convection over Germany in the 
afternoon of the 31th May 2020. This was expected because GFS often shows too high 
low-level moisture values over Central Europe, especially in spring and summer.

GFS: https://www.meteoprime.ch/wrf_gfs/20200531_00/
ICON-EU: https://www.meteoprime.ch/wrf_icon/20200531_00/

PR #220, opened 2023-03-31
Support for dwd suite of models

WQadam:
Addition of DWD suite of models, specifically the ICON models.

1.New Vtables created for icon pressure levels and models levels
2.ReadMe.icon added to add information about the running and downloading specifics needed to run the icon data through WPS.
3.Addition of DWD identifier in rd_grib2.f and rd_grib1.f, plus the addition of level code 150 in rd_grib2.f as icon-eu model levels are on generalized vertical height coordinates. 
4.Addition of new soil levels for moisture and temperatures, plus WW and GEOPT as an optional variable to be processed by metgrid as this is available from the ICON suite of models. File changed for this is METGRID.TBL.ARW
5. rrpr.F changed to add conversion of icon soil moisture to correct units from kg m-2 to m3 m-2 based on if level 0 is present and the model is DWD from the grib file.

Data from PR154 https://github.com/wrf-model/WPS/pull/154 should work and be sufficient for testing.
If any other testing is needed i can provide sample run scripts.

### DIFF
--- a/ungrib/src/rrpr.F
+++ b/ungrib/src/rrpr.F
@@ -1484,7 +1484,7 @@ subroutine fix_icon_soilm (ix, jx)
   call get_storage(200100, 'SOILM486', soilm486, ix, jx)
   call get_storage(200100, 'SOILM999', soilm999, ix, jx)
     !convert kg m-2 to m3 m-3
-    soilm001 = soilm000 / 10.
+    soilm000 = soilm000 / 10.
     soilm002 = soilm002 / 20.
     soilm006 = soilm006 / 60.
     soilm018 = soilm018 / 180.
--- a/ungrib/src/rrpr.F
+++ b/ungrib/src/rrpr.F
@@ -789,7 +789,16 @@ subroutine rrpr(hstart, ntimes, interval, nlvl, maxlvl, plvl, &
               call fix_ruc_soilm (map%nx, map%ny)
             endif
        endif
-
+! Convert soil moisture in ICON input from kg m-2 to m3 m-3      
+       if (index(map%source,'DWD') .ne. 0) then
+          if (is_there(200100, 'SOILM000')) then
+           call get_dims(200100, 'SOILM000')
+              print *,'Adjusting ICON soil moisture'
+           call mprintf(.true.,DEBUG, &
+              "RRPR:   Adjusting ICON soil moisture ")
+           call fix_icon_soilm (map%nx, map%ny)           
+          endif
+       endif
 !
 ! Check to see if we need to fill SOILHGT from SOILGEO.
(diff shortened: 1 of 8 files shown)


weiwangncar:
@WQadam Thanks for putting this PR together too. I was also able to run through your code, after correcting two errors in rrpr.F and rd_grib2.F. I also noticed that there are some differences between your Vtable and the one in PR#221.

WQadam:
> @WQadam Thanks for putting this PR together too. I was also able to run through your code, after correcting two errors in rrpr.F and rd_grib2.F. I also noticed that there are some differences between your Vtable and the one in PR#221.

Oops yeah that my bad with the brackets, should have doubled checked with the copy and paste from my other repository. In the rrpr file was that to do with the differing soil levels, ie 000 to 001? 

In terms of the differing VTables i notice PR#221 has HGT in the model level table which doesn't exist on the open data server, you can only download HGT for pressure levels. Im not sure if this is available on the archive server mentioned in my readme as i have never had access to this but some people who have asked for my code to use have said they are using this server.

For the differing soil levels ie 000 or 001 then these are filled by the same data according to the icon user guide and what i have put in my readme so which level you use is a personal choice and shouldn't make a difference in the long run. 
i think the only other difference is W which is available for model levels and thought that might be a good edition so the model can have an additional startup variable to perhaps reduce spin up time for convection, but i haven't got any tests to prove this as of yet.

And in response to question 4 on PR#221 it should be specific humidity in kg kg-1

hope that helps

ungrib/src/rrpr.F
```diff
@@ -1451,3 +1460,56 @@ subroutine fix_gefs_landmask (ix, jx, plvl)
...
+  call get_storage(200100, 'SOILM000', soilm000, ix, jx)
+  call get_storage(200100, 'SOILM002', soilm002, ix, jx)
+  call get_storage(200100, 'SOILM006', soilm006, ix, jx)
+  call get_storage(200100, 'SOILM018', soilm018, ix, jx)
+  call get_storage(200100, 'SOILM054', soilm054, ix, jx)
+  call get_storage(200100, 'SOILM162', soilm162, ix, jx)
+  call get_storage(200100, 'SOILM486', soilm486, ix, jx)
+  call get_storage(200100, 'SOILM999', soilm999, ix, jx)
+    !convert kg m-2 to m3 m-3
+    soilm001 = soilm000 / 10.
```
sael9740, line 1487: @WQadam - Did you mean to use `soilm001` here? I imagine you meant to use `soilm000`?

weiwangncar:
@WQadam Thanks for your effort in improving the PR#221, and we appreciate it.

PR #222, opened 2023-04-13
added grib2 table parameter codes according to new ECMWF open data and grib1 geopotential height

bakamotokatas:
grib2 table parameter codes according to https://data.ecmwf.int/forecasts/ dataset in the link is added.

ECMWF made some of the variables open in 0.4 resolution. But unfortunately, it does not cover all the variables required to run WRF but can be mixed with other datasets.

also, the grib1 param code for geopotential height is added.

### DIFF
--- a/ungrib/Variable_Tables/Vtable.ECMWF
+++ b/ungrib/Variable_Tables/Vtable.ECMWF
@@ -12,7 +12,7 @@ Param| Type |Level1|Level2| Name     | Units   | Description                    
  167 |  1   |   0  |      | TT       | K        | Temperature          At  2 m             |  0  |  0  |  0  | 103 |
  168 |  1   |   0  |      | DEWPT    | K        |                                          |  0  |  0  |     | 103 |
      |  1   |   0  |      | RH       | %        | Relative Humidity    At  2 m             |  0  |  0  |     | 103 | 
- 172 |  1   |   0  |      | LANDSEA  | 0/1 Flag | Land/Sea flag                            |  0  |  0  |     | 103 | 
+ 172 |  1   |   0  |      | LANDSEA  | 0/1 Flag | Land/Sea flag                            |  2  |  0  |  0  |  1  | 
  129 |  1   |   0  |      | SOILGEO  | m2 s-2   |                                          |  0  |  0  |     | 103 | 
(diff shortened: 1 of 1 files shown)


weiwangncar:
@bakamotokatas Is it possible to provide a test dataset to try this Vtable? You can also demonstrate the success by providing an ungrib output log. Thanks.

bakamotokatas:
I've also added the land-sea mask and soil temperature (0-7cm) to Vtable now.

For the grib2 part, I've included the ungrib.log and g2print.exe results. Since the data is 30MB with bz compression, I couldn't upload it here. You can download it from the link I uploaded: https://filebin.net/ymyd406fxwqhgbql or access the original website at [https://data.ecmwf.int/ecpds/home/opendata/${date}/${baserun}z/0p4-beta/oper/](https://data.ecmwf.int/ecpds/home/opendata/$%7Bdate%7D/$%7Bbaserun%7Dz/0p4-beta/oper/) (the file ends with grib2).

I've also included the g1print.exe log and ungrib.log for geopotential height in the operational data. However, I'm unsure if I'm allowed to share this data. It is the same as the geopotential height value 156 mentioned in this link: https://www.ecmwf.int/en/forecasts/datasets/open-data

[ecmwf_opendata_example_ungriblog_and_g2printout.zip](https://github.com/wrf-model/WPS/files/11229980/ecmwf_opendata_example_ungriblog_and_g2printout.zip)
[ecmwf_grib1_geopotential_height.zip](https://github.com/wrf-model/WPS/files/11229982/ecmwf_grib1_geopotential_height.zip)

jimbresch:
I think the addition of the grib2 codes are fine and they appear to be correct with the caveat that the EC grib2 files don't contain enough fields to initialize WRF (namely the land surface fields). I don't particularly like having so many grib2 parameter entries blank, but it does indicate which fields are not present in the files.

The rrpr code will use HGT when available, otherwise it will compute it using GEOPT. So, the modified Vtable should work with the NCAR/RDA archived files.

jimbresch (approved):

weiwangncar (approved):

weiwangncar:
In some of the EC data, the surface fields are in grib1 format.

weiwangncar:
@bakamotokatas Do you have trouble with recent EC grib2 dataset? In particular, the soil data for temperature and moisture.

PR #223, opened 2023-04-20
No need to write GEOPT and SPECHUMD in intermediate file

weiwangncar:
In the two newly added Vtables for ICON model and pressure level data, we remove the field that is not needed in the intermediate file. In the Vtable for pressure level data, GEOPT is not needed since HGT is calculated in ungrib. In the Vtable for model level data, the SPECHUMD is no longer needed since RH has been created.


weiwangncar:
@mgduda @jimbresch Do you think we can include this in the release?

mgduda (approved):

Issue #240, opened 2023-12-22
Intel LLVM Compilers fail to build WPSv4.5


HathewayWill:
Simlar to this issue:
https://github.com/wrf-model/WRF/issues/1898
https://github.com/wrf-model/WRF/issues/1884

and the solution for WRF in PR
https://github.com/wrf-model/WRF/pull/1893

WPS fails to build with intel LLVM compilers.

I have tried to use intel classic compliers for WPS with WRF 4.5.2 installed with intel LLVM compilers and they are not compatible.

HathewayWill:
@weiwangncar @kkeene44 

[LLVM_WPS_Working.tar.gz](https://github.com/wrf-model/WPS/files/13778714/LLVM_WPS_Working.tar.gz)

I was able to build the WPS programs with intel oneapi llvm with the following changes to the configure.defaults for WPS.  

Can you please verify that they work on your systems?

I will now test to make sure all the programs run as expected.

prashantkul:
Hi @HathewayWill 

I'm running into issues compiling WPS with Intel LLVM compiler. Hoping you can help me.

Attached my compile.log and configure.wps.

my env file is attached as well.

[wps_com.zip](https://github.com/wrf-model/WPS/files/14927094/wps_com.zip)

HathewayWill:
@prashantkul 

I would recommend going to the WRF-MPAS forum https://forum.mmm.ucar.edu/

That's the best place to get help since you are using spack

PR #243, opened 2024-01-11
New intel llvm compilers for WPS

HathewayWill:
This fixes the issue addressed in #240


arch/configure.defaults
```diff
@@ -798,6 +798,30 @@ CPPFLAGS            = -D_UNDERSCORE -DBYTESWAP -DLINUX -DIO_NETCDF -DIO_BINARY -
...
+DM_CC               = mpicc -cc=$(SCC)
+FC                  = CONFIGURE_FC
+CC                  = CONFIGURE_CC
+LD                  = $(FC)
+FFLAGS              = -FR -convert big_endian
+F77FLAGS            = -FI -convert big_endian
+FCSUFFIX            = 
+FNGFLAGS            = $(FFLAGS)
+LDFLAGS             = 
+CFLAGS              = -w -Wno-implicit-int
```
islas, line 819: We should remove these flags as they suppress errors that should be fixed. #245 addresses the syntax issues

islas (changes requested):

arch/configure.defaults
HathewayWill: Okay

HathewayWill:
@islas 

I think I did it sucessfully

arch/configure.defaults
HathewayWill: I think i got it done now

HathewayWill:
> @islas
> 
> I think I did it sucessfully

@weiwangncar @islas 

Is there a release number for the next version of wps for a git clone?

HathewayWill:
@islas @weiwangncar @mgduda @kkeene44 

I think I may have merged somehow.  Can you double check?  I made the changes islas asked for but I think it merged in without approval.

arch/configure.defaults
```diff
@@ -798,6 +798,30 @@ CPPFLAGS            = -D_UNDERSCORE -DBYTESWAP -DLINUX -DIO_NETCDF -DIO_BINARY -
...
 
+########################################################################################################################
+#ARCH    Linux x86_64, Intel oneAPI compiler with icx   # serial serial_NO_GRIB2 dmpar dmpar_NO_GRIB2
+#
+COMPRESSION_LIBS    = CONFIGURE_COMP_L
+COMPRESSION_INC     = CONFIGURE_COMP_I
+FDEFS               = CONFIGURE_FDEFS
+SFC                 = ifx
+SCC                 = icx
+DM_FC               = mpif90 -f90=$(SFC)
```
mgduda, line 809: If there's not a good reason for doing so, I think it would be best to not add `-f90=$(SFC)` (and similarly for `-cc=$(SCC)`) as an argument, as I'm not sure this is a standard argument that's accepted by all MPI wrappers.

Also, I would propose switching from `mpif90` to `mpifort`.

mgduda (changes requested):
Besides the comment on the `-f90` option that I've left, I would suggest moving the new stanza just below the "generic" Intel / x86_64 stanza rather than placing it at the end of the `configure.defaults` file. That way, all of the Intel options would be grouped together.

mgduda:
@HathewayWill I don't think using `mpiifort` is what we're after. My understanding is that `mpiifort` is the MPI Fortran compiler wrapper for Intel's MPI library, while I believe we're trying to make a oneAPI configuration stanza that uses the Intel oneAPI compilers (ifx, icx, etc.) while using more generic MPI compiler wrappers.

HathewayWill:
If I understand it correctly if you call mpifort with the intel mpi
compilers it uses gnu mpifort.  I'll have to test

PR #245, opened 2024-03-14
Syntax fixes for intel oneAPI icx compiler

islas:
Fix syntax issues in ungrib/src/cio.c to follow ISO C99 syntax rules to allow base compilation with `icx` with no flags disabling warnings/errors


islas:
@hahnd A few more changes needed to be made to use ISO C99 syntax in the function arguments, but these changes should work now with `icx`

mgduda (approved):

PR #248, opened 2024-03-29
Remove standard_wrf_dirs and use wildcard for WRF paths

IncubatorShokuhou:
In this PR, I removed the `standard_wrf_dirs` variable and switched to using regular expressions, so we don't need to modify the WRF directory name or update the `standard_wrf_dirs` variable every time a new version is released.

### DIFF
--- a/configure
+++ b/configure
@@ -192,7 +192,6 @@ if [ $? -eq 0 ] ; then
 fi
 
 wrf_dir="none"
-standard_wrf_dirs="WRF WRF-4.0.3 WRF-4.0.2 WRF-4.0.1 WRF-4.0 WRFV3"
 
 if [ ${nowrf} -eq 0 ]; then
    #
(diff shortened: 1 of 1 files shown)


islas:
`-maxdepth` nor `-regex` are standard to `find`
An alternative could be :
`ls .. | grep -E 'WRFV?(-?[0-9]+(\.[0-9](\.[0-9])?)?)?'`

However, caution should be used as both options do not appear in version order. The naming is also variable enough that POSIX `sort` cannot work well to put these in order.

IncubatorShokuhou:
> `-maxdepth` nor `-regex` are standard to `find` An alternative could be : `ls .. | grep -E 'WRFV?(-?[0-9]+(\.[0-9](\.[0-9])?)?)?'`
> 
> However, caution should be used as both options do not appear in version order. The naming is also variable enough that POSIX `sort` cannot work well to put these in order.

I have abandoned the use of complex regular expressions and have instead opted for more explicit if statements. This should be more convenient for maintenance.

islas:
`-V` is part of GNU sort, so I would recommend something like:
`sort  -t. -k 1,1.4n  -k 2,2nr  -k 3,3nr` 
to get the desired effect. The breakout into explicit if statements does look a lot more manageable.

mgduda:
> `-V` is part of GNU sort, so I would recommend something like: `sort -t. -k 1,1.4n -k 2,2nr -k 3,3nr` to get the desired effect. The breakout into explicit if statements does look a lot more manageable.

Should this be
```
sort -t. -k 1.4,1n -k 2,2nr -k 3,3nr
```
instead?

islas:
You're right, I confused the first and last character placement

mgduda:
I'm not sure whether the incorporation of the `-` character in the first key is too clever or not. So an alternative might be
```
sort -t. -k 1.5,1nr -k 2,2nr -k 3,3nr
```
This may be a bit clearer, since all keys are being sorted in reverse numerical order.

configure
```diff
@@ -201,14 +200,25 @@ if [ ${nowrf} -eq 0 ]; then
...
-         if [ -e ../${d}/external/io_netcdf/libwrfio_nf.a ]; then
-            echo "Found what looks like a valid WRF I/O library in ../${d}"
-            wrf_dir="../${d}"
+      # find WRF dir
+      if [ -e "../WRF/external/io_netcdf/libwrfio_nf.a" ]; then
+        echo "Found what looks like a valid WRF I/O library in ../WRF"
+        WRF_DIR="../WRF"
+      else
+        # If the WRF directory does not exist or libwrfio_nf.a is not found, search for directories starting with WRF-4 in descending order of version number
+        for dir in $(ls .. | grep '^WRF-4' | sort -rV); do
```
mgduda, line 209: Following the discussion on this PR, I'd suggest we change `sort -rV` to `sort -t. -k 1.5,1nr -k 2,2nr -k 3,3nr`.

configure
```diff
@@ -201,14 +200,25 @@ if [ ${nowrf} -eq 0 ]; then
...
    if [ -z "$WRF_DIR" ]; then
-   #   for d in WRF WRF-4.0.3 WRF-4.0.2 WRF-4.0.1 WRF-4.0 WRFV3; do
-      for d in ${standard_wrf_dirs}; do
-         if [ -e ../${d}/external/io_netcdf/libwrfio_nf.a ]; then
-            echo "Found what looks like a valid WRF I/O library in ../${d}"
-            wrf_dir="../${d}"
+      # find WRF dir
+      if [ -e "../WRF/external/io_netcdf/libwrfio_nf.a" ]; then
+        echo "Found what looks like a valid WRF I/O library in ../WRF"
+        WRF_DIR="../WRF"
```
mgduda, line 206: Here, as well as on lines 212 and 219, I think `WRF_DIR` should be `wrf_dir`; otherwise, the check on line 235 will fail.

mgduda (changes requested):

PR #250, opened 2024-05-09
Add configuration stanza for the Intel oneAPI compilers with generic MPI wrappers

mgduda:
This PR adds a new configuration stanza for the Intel oneAPI compilers, with compiler wrapper names appropriate for most MPI implementations.

Also included in this PR is an update to the logic in the `configure` script to print compiler version information for the `ifx` compiler.


mgduda:
@islas One additional change that could be included in this PR would rename the existing Intel stanzas from `Intel compiler` to `Intel Classic compilers` just to make the distinction between the old compilers and the newer oneAPI compilers more obvious. Any thoughts?

islas:
Agreed, especially as that is the nomenclature that Intel uses.

I saw the discussion regarding `mpif90` vs `mpifort` - when loading the Intel env via `setvars.sh` (their recommended way of loading) no `mpifort` exists. While using `mpif90` would use Intel's MPI library, I would also expect a DM configuration to work. Not sure what the solution is, but I think it's worth noting.

islas:
Right, I think as long as the MPI implementation you want is first in your path it should work, assuming that implementation supplies an `mpif90` wrapper.

mgduda:
> Right, I think as long as the MPI implementation you want is first in your path it should work, assuming that implementation supplies an `mpif90` wrapper.

Ok, I've just pushed another commit to switch from `mpifort` to `mpif90`.

islas:
I am unfortunately seeing the same behavior @HathewayWill described where without specifying the serial compiler you run the risk of getting something else:

```
/home/aislas/dependencies/mpich/bin/mpif90 -fc=ifx -V
Intel(R) Fortran Compiler for applications running on Intel(R) 64, Version 2024.0.2 Build 20231213
Copyright (C) 1985-2023 Intel Corporation. All rights reserved.

GNU ld version 2.35.2-42.el9
ld: /home/aislas/intel/oneapi/compiler/2024.0/lib/for_main.o: in function `main':
for_main.c:(.text+0x19): undefined reference to `MAIN__'


/home/aislas/dependencies/mpich/bin/mpif90 -V
gfortran: error: unrecognized command-line option ‘-V’


echo $PATH
/home/aislas/tools/ninja/run/bin/:/home/aislas/intel/oneapi/vtune/2024.0/bin64:/home/aislas/intel/oneapi/mpi/2021.11/opt/mpi/libfabric/bin:/home/aislas/intel/oneapi/mpi/2021.11/bin:/home/aislas/intel/oneapi/mkl/2024.0/bin/:/home/aislas/intel/oneapi/itac/2022.0/bin:/home/aislas/intel/oneapi/inspector/2024.0/bin64:/home/aislas/intel/oneapi/dpcpp-ct/2024.0/bin:/home/aislas/intel/oneapi/dev-utilities/2024.0/bin:/home/aislas/intel/oneapi/debugger/2024.0/opt/debugger/bin:/home/aislas/intel/oneapi/compiler/2024.0/opt/oclfpga/bin:/home/aislas/intel/oneapi/compiler/2024.0/bin:/home/aislas/intel/oneapi/advisor/2024.0/bin64:/home/aislas/dependencies/bin:/glade/u/home/aislas/bin:/glade/u/home/aislas/.local/bin/:/glade/u/home/aislas/tools/built/bin:/home/aislas/bin/:/home/aislas/bin/VSCode-linux-x64:/home/aislas/bin/go/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin
```

My path specifically puts the Intel compiler before the system gfortran...

islas:
The WRF stanza does have those args used, for better or worse.

mgduda:
I don't think the `-fc` and `-cc` arguments are standard, which is why I'd prefer to not include them.

@islas Out of curiosity, does the following subroutine compile with both `mpif90 -fc=ifx -c <filename>` and `mpif90 -fc=gfortran -c <filename>` on your system?
```fortran
subroutine nvsend()

   use mpi_f08

   real, dimension(100) :: buf
   integer :: count, dest, tag
   type(MPI_Datatype) :: datatype
   type(MPI_Comm) :: comm
   type(MPI_Request) :: request
   integer :: ierror

   call MPI_Isend(buf, count, datatype, dest, tag, comm, request, ierror)

end subroutine nvsend
```
(Specifically, you should not load a different set of modules between attempts.)

My suspicion is that both won't work, as the `mpi_f08` module was only built with the compiler that was used to build MPICH. This example would illustrate why `-f90=` can't work in general. But perhaps I'm mistaken.

mgduda:
> This also would apply to the wrf implementation @islas @mgduda ??

How we invoke the MPI compiler wrappers should apply equally to WRF and to the WPS.

islas:
@mgduda Suspicion confirmed:
```
[mmm-mystic {127} test/mpi ]$ which mpif90 
/home/aislas/intel/oneapi/mpi/2021.11/bin/mpif90
[mmm-mystic {0} test/mpi ]$ mpif90 -fc=ifx -c simple.f90 
[mmm-mystic {0} test/mpi ]$ ll
total 8
drwxr-xr-x. 2 aislas aislas   40 May 10 10:53 ./
drwxr-xr-x. 4 aislas aislas   82 May 10 10:50 ../
-rw-r--r--. 1 aislas aislas  309 May 10 10:51 simple.f90
-rw-r--r--. 1 aislas aislas 1368 May 10 10:53 simple.o
[mmm-mystic {0} test/mpi ]$ mpif90 -fc=gfortran -c simple.f90 
f951: Fatal Error: Reading module ‘/home/aislas/intel/oneapi/mpi/2021.11/include/mpi/mpi_f08.mod’ at line 1 column 2: Unexpected EOF
compilation terminated.
[mmm-mystic {1} test/mpi ]$ pathmunge ~/dependencies/mpich/bin/
[mmm-mystic {0} test/mpi ]$ which mpif90 
/home/aislas/dependencies/mpich/bin/mpif90
[mmm-mystic {0} test/mpi ]$ mpif90 -fc=ifx -c simple.f90 
simple.f90(3): error #7013: This module file was not generated by any release of this compiler.   [MPI_F08]
   use mpi_f08
-------^
simple.f90(7): error #6457: This derived type name has not been declared.   [MPI_DATATYPE]
   type(MPI_Datatype) :: datatype
--------^
simple.f90(8): error #6457: This derived type name has not been declared.   [MPI_COMM]
   type(MPI_Comm) :: comm
--------^
simple.f90(9): error #6457: This derived type name has not been declared.   [MPI_REQUEST]
   type(MPI_Request) :: request
--------^
compilation aborted for simple.f90 (code 1)
[mmm-mystic {1} test/mpi ]$ mpif90 -fc=gfortran -c simple.f90 
```

So would the onus fall on the user / sys admin to ensure a compatible wrapper is supplied?

Also, while it _did_ do what I wanted and expected (use the compiler I said to), the same cannot be said for OpenMPI:
```
 [mmm-mystic {0} test/mpi ]$ pathmunge ~/dependencies/openmpi/bin/
[mmm-mystic {0} test/mpi ]$ which mpif90 
/home/aislas/dependencies/openmpi/bin/mpif90
[mmm-mystic {0} test/mpi ]$ mpif90 -fc=ifx -c simple.f90 
gfortran: error: unrecognized command-line option ‘-fc=ifx’
[mmm-mystic {1} test/mpi ]$ mpif90 -fc=gfortran -c simple.f90 
gfortran: error: unrecognized command-line option ‘-fc=gfortran’
```

mgduda:
@islas I think your test results do illustrate that (1) `-fc=` is not portable (it's not supported by all MPI compiler wrappers) and (2) `-fc=` isn't generally viable (when using any MPI Fortran modules, one cannot just switch Fortran compilers and expect to use the rest of the MPI library compiled with different compiler).

I would say that it's the responsibility of the sys admin or user to ensure that the available serial Fortran compiler agrees with the `mpif90` wrapper in the user's path.

islas:
Sounds good, I'll make a note to change the WRF stanza in the next update to follow suit.

islas (approved):

mgduda:
@HathewayWill Thanks for the info. On the WPS side, don't think this will affect us, as we don't currently have any configuration stanzas that use the Intel MPI wrappers (`mpiifort`, `mpiicc`, etc.).

PR #255, opened 2024-06-11
When extracting FFLAGS in configure script, expand $(FORMAT_FREE) Make variable

mgduda:
This PR adds logic to expand the `$(FORMAT_FREE)` Make variable when extracting `FFLAGS` from the `configure.wps` file in the `configure` script.

With commit 48f7904d, Fortran free-format flags were moved from `FFLAGS` into a separate Make variable named `FORMAT_FREE` in the `arch/configure.defaults` file. Because the `configure` script extracts `FFLAGS` from the `configure.wps` file when compiling test programs, the `$(FORMAT_FREE)` variable is not expanded, leading to compilation errors for test programs.

This PR introduces extra logic in the configure script to expand `$(FORMAT_FREE)` in the `FFLAGS` string for the specification of the `FORMAT_FREE` variable in the `configure.wps` file, allowing for a full set of valid compiler flags to be used when compiling test programs.

Note: If `FORMAT_FREE` itself contains Make variables, these will not be recursively expanded. Also, `$(FORMAT_FREE)` is the only Make variable that is expanded in `FFLAGS` at present.

### DIFF
--- a/configure
+++ b/configure
@@ -427,6 +427,8 @@ cat > fort_netcdf.f <<EOF
         end program
 EOF
     FFLAGS=`grep ^FFLAGS configure.wps | cut -d"=" -f2-`
+    FORMAT_FREE=`grep ^FORMAT_FREE configure.wps | cut -d"=" -f2-`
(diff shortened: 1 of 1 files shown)


islas (approved):
Works for both make and cmake configuration with gnu

PR #258, opened 2024-09-09
Add aarch64 config defaults

veech:
Add Linux aarch64 GNU compiler architecture to WPS

This PR adds support for Linux aarch64 systems using GNU compilers (gfortran/gcc) to the WPS configure.defaults file. The new architecture definition enables building WPS on ARM64-based Linux systems, supporting both serial and distributed memory parallel (dmpar) configurations, with and without GRIB2 support.

Key features:
- Uses gfortran and gcc compilers
- Configures appropriate flags for big-endian conversion and record markers
- Supports both serial and parallel builds
- Compatible with existing WPS configuration process

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -930,3 +930,31 @@ FORMAT_FIXED        =
 #ARFLAGS            =
 #RANLIB              = 
 #CC_TOOLS            = 
+
+########################################################################################################################
+#ARCH    Linux aarch64, GNU compilers    # serial serial_NO_GRIB2 dmpar dmpar_NO_GRIB2
+#
+COMPRESSION_LIBS    = CONFIGURE_COMP_L
+COMPRESSION_INC     = CONFIGURE_COMP_I
+FDEFS               = CONFIGURE_FDEFS
+SFC                 = gfortran
+SCC                 = gcc
+DM_FC               = mpif90
+DM_CC               = mpicc
+FC                  = CONFIGURE_FC
+CC                  = CONFIGURE_CC
+LD                  = $(FC)
+FFLAGS              = $(FORMAT_FREE) -O -fconvert=big-endian -frecord-marker=4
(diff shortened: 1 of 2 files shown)


mgduda:
@veech I may be missing something, but I think the existing `Linux x86_64, gfortran` stanza in `configure.defaults` should work equally for `x86_64` and `aarch64`. I'd suggest creating a branch from `master` or the `v4.6.0` tag and simply generalizing the name of the `Linux x86_64, gfortran` stanza to something like `Linux, GNU compilers`. But in any case, creating a PR from a branch and not from the head of your `master` branch would be good.

veech:
For a little more context, I'm running the build within a Docker container on my Apple Silicon Mac. When running `configure` without the lines I added, I get this:
```
Will use NETCDF in dir: /wrf/LIBRARIES/netcdf
Found what looks like a valid WRF I/O library in ../WRF
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
Found Jasper environment variables for GRIB2 support...
  $JASPERLIB = /wrf/LIBRARIES/grib2/lib
  $JASPERINC = /wrf/LIBRARIES/grib2/include
------------------------------------------------------------------------
Please select from among the following supported platforms.

Enter selection [1-0] : 
```

When changing the line in `configure.defaults` to:
```
#ARCH    Linux, GNU compilers    # serial serial_NO_GRIB2 dmpar dmpar_NO_GRIB2
```

I get the same as above.

Sorry about the PR from master, what's the best way to fix this? Should I create a new PR from a branch?

mgduda:
@veech Apologies for the long delay in following up. It looks like the change that have been proposed in PR #262 should allow you to build with the GNU compilers on Linux aarch64 systems. Would you be willing to see whether those changes work for you?

PR #262, opened 2024-10-10
Update configure.defaults to support a Linux aarch64 gfortran option

benkirk:
This one line change extends the  `Linux x86_64, gfortran` stanza to support `aarch64` (ARM processors) as well by simply changing the definition to `Linux x86_64 aarch64, gfortran` 

Since there are no required differences when using `gfortran` on these two architectures this seemed like the best approach.

Let me know if you would instead prefer a new stanza, with identical content.

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -189,7 +189,7 @@ CPPFLAGS            = -D_UNDERSCORE -DBYTESWAP -DLINUX -DIO_NETCDF -DBIT32 -DNO_
 RANLIB              = ranlib
 
 ########################################################################################################################
-#ARCH    Linux x86_64, gfortran   # serial serial_NO_GRIB2 dmpar dmpar_NO_GRIB2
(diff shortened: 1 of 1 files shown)


joehellmersNOAA:
Make this change worked for the configure on Linux aarch64 using GFortran on Ubuntu 24.04 server.

HathewayWill:
@islas @mgduda @kkeene44 

I can confirm making this change allows for WPS 4.6.0 to show a configure option and build successfully (ungrib/metgrid/geogrid)
```

 workhorse@workhorsearm64:~/WRFCHEM/WPS-4.6.0$ ls *.exe
geogrid.exe  metgrid.exe  ungrib.exe
```

mgduda:
@benkirk If you don't mind, I'd like to force-push a change to your commit message to remove the sentence "Let me know if you would instead prefer a new stanza, with identical content." and to wrap the lines at 80 characters.

benkirk:
Absolutely no problem @mgduda !

Issue #264, opened 2024-11-27
Change in soil variables from ECMWF Cycle 49r1


gabrielcassol:
On November 12, ECMWF released version 49r1 of the operational IFS. Despite all the improvements, they modified the soil variables (temperature and moisture) which broke the [ECMWF Vtable](https://github.com/wrf-model/WPS/blob/master/ungrib/Variable_Tables/Vtable.ECMWF) that had been released in the 4.6.0 version of the WPS.

I contacted the ECMWF team and they gave me this feedback. It is necessary to make a change within Ungrib, more specifically in the [rd_grib2.F](https://github.com/wrf-model/WPS/blob/335c76a111f84503e8b963abaf273ea8053645bb/ungrib/src/rd_grib2.F) code so that it works with the new IFS soil variables.

With 49R1 cycle, encoding of soild parameters changed

- from [stl1](https://codes.ecmwf.int/grib/param-db/139) , [stl2](https://codes.ecmwf.int/grib/param-db/170), [stl3](https://codes.ecmwf.int/grib/param-db/183) , and [stl4 ](https://codes.ecmwf.int/grib/param-db/236)changed to [sot](https://codes.ecmwf.int/grib/param-db/260360).
- and from [swvl1](https://codes.ecmwf.int/grib/param-db/39) , [swvl2](https://codes.ecmwf.int/grib/param-db/40), [swvl3](https://codes.ecmwf.int/grib/param-db/41), and [swvl4](https://codes.ecmwf.int/grib/param-db/42) changed to [vsw](https://codes.ecmwf.int/grib/param-db/260199)

![image](https://github.com/user-attachments/assets/27eaffa4-5c7a-4ef9-9a34-3b0067b55147)

ECMWF is currently one of the best global weather forecasting models. It is essential to have this data working in WRF so that regional weather centers can run WRF for their area of ​​interest. Could you take a look and try to find a solution to this problem?

Best Regards,
Gabriel

weiwangncar:
@jimbresch Take a look at this post.

HathewayWill:
https://forum.mmm.ucar.edu/threads/change-in-soil-variables-from-ecmwf-cycle-49r1.19864/#post-48713


@weiwangncar @jimbresch

HathewayWill:
@weiwangncar @mgduda @kkeene44 

https://forum.mmm.ucar.edu/threads/change-in-soil-variables-from-ecmwf-cycle-49r1.19864/post-52211

[...]

PR #265, opened 2024-12-20
New ERA5 Vtable

JoshLovesFun:
I have made a new ERA5 Vtable using the Vtable.ECMWF as a template. I have not made any changes that should make this Vtable.ERA5 function differently from the current Vtable.ECMWF. I have just made some clarifications and added some comments.

This Vtable.ERA5 is needed, because it will diverge from the Vtable.ECMWF when the Vtable.ECMWF is updated based on the new changes to ECMWF: #264 
I have tested it, and it worked successfully. Attached are my Ungrib logs.
[ungrib_pres.log](https://github.com/user-attachments/files/18214720/ungrib_pres.log)
[ungrib_sfc.log](https://github.com/user-attachments/files/18214722/ungrib_sfc.log)


ungrib/Variable_Tables/README
```diff
@@ -21,6 +21,8 @@ Vtable.ECMWF_sigma         ECMWF sigma-level output
 Vtable.ERA-interim.ml           ERA interim model level
 Vtable.ERA-interim.pl           ERA interim pressure level output
 
+Vtable.ERA5                ERA5 single level and pressure level data from the Climate Data Store
```
weiwangncar: @JoshLovesFun Do you mean 'sigma' level?

JoshLovesFun: @weiwangncar In this case, it is actually pressure levels. Because the specific dataset that this Vtable is targeting is the ERA5 data in the climate data store that has already been regridded to pressure levels from the native hybrid-sigma levels. So thinking about it more, this Vtable should be called "Vtable.ERA5.pl".

If I work with the native ERA5 model data in the future, I could make a "Vtable.ERA5.native"

I think the native ERA5 data is a bit more difficult to download and use, but I haven't tried it. I think most user's would get the pressure level data.

HathewayWill:
I was able to use the existing ECMWF table with the new changes and it worked for me, not sure why it happened though @JoshLovesFun

JoshLovesFun:
> I was able to use the existing ECMWF table with the new changes and it worked for me, not sure why it happened though @JoshLovesFun

Both this table and the existing table should effectively be identical for ERA5 data. I would think neither table would work as we want it to work for the new ECMWF version though.

HathewayWill:
> > I was able to use the existing ECMWF table with the new changes and it worked for me, not sure why it happened though @JoshLovesFun
> 
> Both this table and the existing table should effectively be identical for ERA5 data. I would think neither table would work as we want it to work for the new ECMWF version though.

that makes more sense thank you for clarifying

ungrib/Variable_Tables/Vtable.ERA5.pl
```diff
@@ -0,0 +1,83 @@
...
+#  34 = Sea surface temperature
+# 141 = Snow depth
+# 139 = Soil temperature level 1
+# 170 = Soil temperature level 2
+# 183 = Soil temperature level 3
+# 236 = Soil temperature level 4
+#  39 = Volumetric soil water layer 1
+#  40 = Volumetric soil water layer 2
+#  41 = Volumetric soil water layer 3
+#  42 = Volumetric soil water layer 4
```
mgduda: Could you add an EOL character to the end of this file?

mgduda (changes requested):

HathewayWill:
I think that the current ecmwf and this new table need to have different file names in ungrib or be apart of one massive ecmwf ungrib file.  Just my thoughts what do you think?

ungrib/Variable_Tables/Vtable.ERA5.pl
JoshLovesFun: > Could you add an EOL character to the end of this file?

I will do this when I get a chance. Hopefully soon.

JoshLovesFun: @mgduda done

PR #269, opened 2025-04-12
Only adjust GFS RH in rrpr prior to V16 (3/22/21)

jimbresch:
NCEP changed the GFS RH output in V16 to be with respect to water rather than a function of ice and water, so we no longer need to adjust its RH in ungrib. This GFS change occurred on March 22, 2021 and there was no mention of it (or the UPP) in the Service Change Notice for V16. Tests of the modified rrpr show it correctly processes the V16 RH.  Without this mod, the humidity at temperatures below 0 C is too low. GEFS output was also tested and included in this mod.


jimbresch:
Tests show that the GFS RH still needs adjustment, so this change is not needed.

PR #275, opened 2025-06-12
Change glevel 1 and 2 to ints to prevent floating point equality checks

islas:
All comparisons to glevel1 and glevel2 are against integer values, relying on implicit conversion during the check if done somewhat correctly. When implicit conversion is not done, the floating point to integer .eq. check can easily fail for values that should normally succeed.

Whilst the calculations to generate glevel1 and glevel2 use floating point math, assigning into an integer variable guarantees a float to int conversion is done before the comparison, thus preventing instances where instead of float to int compilers may other arithmetic operations for comparison such as int to float.

### DIFF
--- a/ungrib/src/rd_grib2.F
+++ b/ungrib/src/rd_grib2.F
@@ -74,7 +74,7 @@
       real , dimension(maxlvl) :: plvl
       integer :: nlvl
       integer , dimension(maxlvl) :: level_array
-      real :: glevel1, glevel2
+      integer :: glevel1, glevel2
       logical :: first = .true.
       real :: pmin
 


islas:
More details can be found in this thread:
https://forum.mmm.ucar.edu/threads/subsoil-level-not-found-in-ungrib-file.21884/post-53661

HathewayWill:
@islas will this be updated?

HathewayWill:
@islas

HathewayWill:
I can confirm this issues also occurs with aocc compilers

ungrib/src/rd_grib2.F
```diff
@@ -74,7 +74,7 @@ SUBROUTINE rd_grib2(junit, gribflnm, hdate,
       real , dimension(maxlvl) :: plvl
       integer :: nlvl
       integer , dimension(maxlvl) :: level_array
-      real :: glevel1, glevel2
+      integer :: glevel1, glevel2
```
mgduda, line 77: Switching `glevel1` and `glevel2` to `integer` values does carry a risk: any floating-point values assigned to them will be _truncated_ to integer values; e.g.,
```fortran
glevel1 = 9.999999
```
will result in `glevel1` containing a value of `9`, when the code in `rd_grib2.F` for comparing levels would probably have assumed that `glevel1` would compare equally with a value of `10`.

One solution may be to explicitly round floating-point values assigned to `glevel1` and `glevel2` to the nearest integer with `nint`. Another solution may be to simply apply `nint` when comparing `glevel1` and `glevel2` with integers, and to otherwise leave these two variables declared as `real` values.

mgduda:
PR #292 addresses the floating-point comparison issue in a slightly different way (by rounding `glevel1` and `glevel2` to the nearest integer rather than truncating their right-hand sides to integer values). Since #292 has been merged, I'm closing this PR.

PR #276, opened 2025-06-12
Add default values to controllable cmake options

islas:
The default options are useful when completely bypassing the configuration prompts. Without these default options certain comparison operators are left evaluating blank expansions that would otherwise lead to malformed syntax


islas:
More details can be found in this forum thread :
https://forum.mmm.ucar.edu/threads/issues-configuring-wps-4-6-0-w-wrf-4-7-0-using-configure_new-solved.22161/#post-53682

mgduda (approved):

PR #277, opened 2025-08-01
Add python domain visualization script based on plotgrids_new.ncl

islas:
This utility script takes the original intent of plotgrids[_new].ncl and rewrites it in python and adds multiple convenience features.

Extra info about domains is plotted next to each domain near their respective ij start (SW corner in the projected coordinates).

An analogous perspective view can be plotted alongside the domain view. This perspective view is taken as a straight down look on the Earth from a geostationary orbit altitude looking directly at `ref_lat`/`ref_lon` This may aid in assessing the "quality" of the projection with respect to distortion and map scale factor.

To add in further diagnostics, a verification flag prompts the loading of the respective geogrid .nc files for the provided namelist and plots a field (default `HGT_M)` to the projections using the mass staggered lat/lon coordinates. This can then be compared to the domain boxes to ensure a match or lack thereof if something has gone amiss.

The following domain checks are performed to ensure good domain layout:
* nested domain size matching `parent_grid_ratio`
* nested domain boundary at least 5 parent cells away from parent boundary
* nested domain overlapping within the same parent

Rotated lat-lon projection is NOT SUPPORTED

Typical usage:
```
./util/plot_domains.py
```

### DIFF
--- a/util/plot_domains.py
+++ b/util/plot_domains.py
@@ -1,0 +1,611 @@
+#!/usr/bin/env python3
+from collections import OrderedDict
+import argparse
+import importlib
+import os
+import re
+
+import matplotlib as mpl
+import matplotlib.patheffects as PathEffects
+import matplotlib.pyplot as plt
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+
+
+def read_nml( filename ):
+  """Read a strict subset of Fortran 90 namelist
+
+  This ONLY reads namelists in the format:
+  &group
+   key0 = value0,
+   key1  =  value0, value1,
+   key2='value0', 'value1',
+   key3  =   "value0", "value1",
+  /
+  ...and so on...
+  Index slices, multiple key-value pairs on the same line,
+  and other features are NOT supported. For full support
+  please utilize f90nml.
+
+  Spaces between key and value and commas does not matter.
+  Multivalue entries MUST be delimited by commas.
+  """
+  contents = ""
+  with open( filename, "r" ) as f:
+    contents = f.read()
+
+  nml = OrderedDict()
+  for match in re.finditer( r"&(?P<group>\w+)(?P<kv_pairs>.*?)\n*^[ ]*/", contents, re.S | re.M ):
+    group = match.group( "group" )
+    nml[group] = OrderedDict()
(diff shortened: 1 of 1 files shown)


islas:
# Example outputs

## Lambert Conformal
### Domain setup
Lambert Conformal over CONUS

### Verify against geogrid
Lambert Conformal over CONUS with geogrid output overlayed

## Mercator
### Domain setup
Mercator over Malaysia

### Verify against geogrid
Mercator over Malaysia with geogrid output overlayed

## Polar Stereographic (north)
### Domain setup
Polar Stereographic over Iceland

### Verify against geogrid
Polar Stereographic over Iceland with geogrid output overlayed

## Polar Stereographic (south)
### Domain setup
Polar Stereographic over Antarctica

### Verify against geogrid
Polar Stereographic over Antarctica with geogrid output overlayed

## Cylindrical Equidistant (lat-lon)
### Domain setup
Lat-Lon over Panama

### Verify against geogrid
Lat-Lon over Panama with geogrid output overlayed

## Output with perspective view
### Lambert Conformal over Vancouver with perspective view
Perspective view is taken as a straight down look on the Earth from a geostationary orbit altitude looking directly at `ref_lat`/`ref_lon`
Screenshot from 2025-08-01 13-02-59

Raw image of one of the outputs for viewing text:
Screenshot from 2025-08-01 13-01-34

mgduda:
@islas This is nice! I have no major comments, other than to suggest reworking the commit history (perhaps a single commit is all that's needed?) since we don't do "squash and merge" for WPS PRs. I'd also suggest rewording the PR title so that it doesn't frame this as work in progress ("initial take"); in my view, this PR adds a new Python script for visualizing domains that may be based on plotgrid_new.ncl, but it contains substantial added functionality.

weiwangncar:
@islas This is cool! Can the fonts for the domain info be changed to something simpler?

islas:
@weiwangncar Do you mean something like so?
black_noborder

I can add it as an option if we think that would be useful (to toggle between the two text styles). I originally opted for the white + black border because while it is at times a little difficult to read it does generally better in all types of backgrounds vs just black text which can become unreadable with any non-light background

weiwangncar:
@islas Yes. I like this!

islas:
@weiwangncar I made the simple text the default and kept the white + black outline as an alternate option

weiwangncar:
@islas Thanks!

mgduda (approved):

PR #281, opened 2026-02-04
Post process geogrid output to compute GWD input fields directly from source data

islas:
These changes calculate the necessary fields in a manner that accounts for subgrid-scale orography. This in in support of subgrid orographic parameterization revisions for GWD in WRF.

Currently the gravity wave drag (GWD) input fields from WPS/geogrid are computed from averaging of the source data then performing the calculations for oa1-oa4, ol1-ol4, var, and con. This utility program instead calculates these orographic statistics directly from the source data at a subgrid-scale assuming the domain dx/dy are larger than the source data extent per value.

The code is designed to be run using the entry point `./util/compute_gwdo.py` and will use `./namelist.wps` if no namelist is provided. 

The namelist is read to determine the geogrid domains output, source data tiles locations, and domain spacing. To avoid recalculating lat/lon positioning, the values are read from the mass staggered lat/lon locations within `geo_em.d01.nc`, `geo_em.d02.nc`, and so on. Source data is loaded as tiles are needed to compute the orographic statistics for a given domain cell based on the dx/dy extent of the cell at that lat/lon. 

Once all calculations are done for all cells within a domain file, that file's original oa1-oa4, ol1-ol4, var, and con are modified with the newly calculated subgrid-scale statistics. Additionally, a `ELVMAX` field is created. This is repeated for all domains listed within the namelist.


util/compute_gwdo.py
```diff
@@ -0,0 +1,181 @@
...
+from collections import OrderedDict
+
+import netCDF4 as nc4
+
+from SourceData import SourceData
+from OrographyStats import OrographyStats
+
+
+# https://stackoverflow.com/a/34325723
+def progerss_bar( iteration, total, prefix="", suffix="", precision=1, length=100, fill="*", empty="-", end="\r"):
```
mgduda: Would it be better to name this function `progress_bar`?

util/compute_gwdo.py
```diff
@@ -0,0 +1,181 @@
...
+  key0 = value0,
+  key1  =  value0, value1,
+  key2='value0', 'value1',
+  key3  =   "value0", "value1",
+  /
+  ...and so on...
+  Index slices, multiple key-value pairs on the same line,
+  and other features are NOT supported. For full support
+  please utilize f90nml.
+  Spaces between key and value and commas does not matter.
```
mgduda: "Spaces" -> "do not matter"

util/OrographyStats.py
```diff
@@ -0,0 +1,95 @@
+import numpy as np
+
+
+class OrographyStats:
+  def __init__( self, box ):
+    self.box  = box
+
+    self.mean = np.mean( self.box )
+    # var (actulally stddev)
```
mgduda: "actulally" -> "actually"

util/SourceData.py
```diff
@@ -0,0 +1,257 @@
...
+
+  def get_dtype( self ):
+    endian = ">" if self.endian == "big" else "<"
+    signed = "i" if self.signed else "u"
+    return f"{endian}{signed}{self.wordsize}"
+
+
+class SourceData:
+  def __init__( self, name, path ):
+    self._earth_radius = 6371229.0
```
mgduda: I think the WPS and WRF use 6370000.0 m as the sphere radius (MPAS uses 6371229.0 m).

util/SourceData.py
```diff
@@ -0,0 +1,257 @@
...
+    self._source_path  = path
+    self._index        = IndexData( self._source_path )
+
+    self._projection   = None
+    self._npts_x       = int( 360.0 / self._index.dx )
+    self._pts_per_deg  = int( 1.0 / self._index.dx )
+    self._subgrid_m_dx = 2.0 * np.pi * self._earth_radius / self._npts_x
+
+    self._tile_data  = TileData(
+                                int( int( 360.0 / self._index.dx ) / self._index.tile_x ),
```
mgduda: Rather than repeating `int( 360.0 / self._index.dx )`, would it be better to use `self._npts_x`?

util/SourceData.py
```diff
@@ -0,0 +1,257 @@
...
+                                int( int( 180.0 / self._index.dy ) / self._index.tile_y ),
```
mgduda: Would it make sense to define `self._npts_y` as we do for `self._npts_x`, and to use that value here?

util/compute_gwdo.py
```diff
@@ -0,0 +1,181 @@
...
+                          ]
+                          )
+                    )
+      nml[group][key] = values if len( values ) > 1 else values[0]
+  return nml
+
+
+def main():
+  parser = argparse.ArgumentParser(
+                                    "./compute_gwdo.py",
```
mgduda: What do you think about removing the `./` prefix from the program name? Since this resides in the `./util` subdirectory, my guess is that most users will run this as `./util/compute_gwdo.py` in practice; but it's possible that users may also have an entirely separate run directory from which they will run this script.

util/compute_gwdo.py
```diff
@@ -0,0 +1,181 @@
...
+                                    )
+  parser.add_argument(
+                      "-n", "--namelist",
+                      help="WPS namelist",
+                      type=str,
+                      default="./namelist.wps"
+                      )
+  parser.add_argument(
+                      "-d", "--dataset",
+                      help="Topographic dataset to use, in geogrid format",
```
mgduda: Should we add more info here (or elsewhere?) to indicate that only global datasets on a lat-lon grid will work? For example, using the SRTM 3s data with `-d SRTM_topo_3s` will fail.

mgduda (changes requested):
I'll take more time to try out domains over different parts of the globe, but for now I thought I'd leave some minor initial comments.

util/compute_gwdo.py
islas: What about something like `<path to script>/compute_gwdo.py`?

mgduda: I think it would be reasonable to assume (expect?) that users have some experience working in a shell, and they will understand that they can either (1) provide a relative or absolute path to the `compute_gwdo.py` program when running it, or (2) add the path to `compute_gwdo.py` to their `PATH` environment variable and subsequently run `compute_gwdo.py` with no path specified.

util/compute_gwdo.py
islas: A quick investigation found that taking a subset will require further work and analysis. For the sake of time I think this is sufficient to at least allow working with the global gmted data and noting it.

util/compute_gwdo.py
```diff
@@ -0,0 +1,183 @@
...
+
+    ol1  = geo_data.variables[ "OL1" ]
+    ol2  = geo_data.variables[ "OL2" ]
+    ol3  = geo_data.variables[ "OL3" ]
+    ol4  = geo_data.variables[ "OL4" ]
+
+    if "MAX_EL" in geo_data.variables:
+      max_el = geo_data.variables["MAX_EL"]
+    else:
+      max_el = geo_data.createVariable( "MAX_EL", "f4", ( var.dimensions ) )
```
mgduda: @weiwangncar Do you know if the `MAX_EL` variable is used in the GWDO scheme in WRF? Or should this name actually be `ELVMAX`?

util/compute_gwdo.py
```diff
@@ -0,0 +1,183 @@
...
+
+  # Scale-awareness, require min box size 4x4 using nominal dx of orographic data
+  sg_delta = topo_source._subgrid_m_dx
+  dc = min( float( nml["geogrid"]["dx"] ), float( nml["geogrid"]["dy"] ) ) * 2
+  hratio = dc / sg_delta
+  if hratio < 4.0:
+    dc = sg_delta * 4.0
+
+  box_size_x  = dc
+  box_size_y  = dc
```
mgduda: If I'm not misreading the code, it looks like the same `box_size_x` and `box_size_y` values will be used for all domains. Instead, I think the box size should be sized according to the nominal grid distance in each domain.

mgduda (changes requested):

util/compute_gwdo.py
weiwangncar: @mgduda As it is defined now, the geogrid (+compute_gwdo.py) outputs MAX_EL, but in the WRF model, it is defined as ELVMAX. That is an error. I made a change in my test code to use MAX_EL in the Registry, but when it is passed to pbl driver, I let it use ELVMAX. So if we would like to keep the variable name the same from geogrid to model, we should change the variable name to ELVMAX. It looks like MPAS will output ELVMAX from init_atmosphere if I read MPAS PR-1408 correctly. In that case, we probably should do that same here. Agree?

mgduda: Agreed -- we'll change the variable name in the `compute_gwdo.py` script to `ELVMAX`. Thanks for double-checking this!

util/compute_gwdo.py
mgduda: I've addressed this issue with commit ae591aa8.

util/compute_gwdo.py
mgduda: I've changed `MAX_EL` to `ELVMAX` with commit d06a9899, and commit 40c393de sets the `units` and `description` attributes for `ELVMAX`.

mgduda (approved):

mgduda:
@weiwangncar Would you be willing to give commits d06a9899, 40c393de, and ae591aa8 a quick look?

weiwangncar:
@mgduda Thanks for your careful testing and review. These changes look good to me. I have tested the code and the output looks good for both parent and nest domains.

weiwangncar:
@mgduda I will give WRF model a try next.

weiwangncar:
@mgduda My model run went well. I will approve this PR now. One question, for a large number of grid points, the python script can take a while. Should we mention it in the PR and/or release notes?

weiwangncar (approved):

mgduda:
> @mgduda My model run went well. I will approve this PR now. One question, for a large number of grid points, the python script can take a while. Should we mention it in the PR and/or release notes?

Since the script does print a progress bar, perhaps that would be enough to give users a quick sense of how long the processing may take to complete. If we do get feedback on the runtime, though, we could follow up with some changes to provide an estimated time to completion, or we could focus some future effort on optimization.

weiwangncar:
@mgduda That's ok with me.

PR #282, opened 2026-04-06
update ungrib to handle files larger than 2GB

guoqing-noaa:
1. Update byte offset variables (start/newpos in C, IB/JB in Fortran, iseek/lskip/KS in callers) to use 64-bit
2. Byte count variables (NB/lgrib/no/size/nactual) remain 32-bit, assuming individual GRIB messages are under 2GB
3. Use `long long` C type to match Fortran's `INTEGER(KIND=8)`

### DIFF
--- a/ungrib/src/g2print.F
+++ b/ungrib/src/g2print.F
@@ -161,8 +161,8 @@ end program g2print
       character(len=8)  :: pabbrev
       character(len=20) :: labbrev
       character(len=80) :: tabbrev
-      integer :: lskip, lgrib
-      integer :: junit, itot, icount, iseek
+      integer(kind=8) :: lskip, iseek
+      integer :: lgrib, junit, itot, icount
       integer :: grib_edition
       integer :: i, j, ireaderr, ith
       integer :: currlen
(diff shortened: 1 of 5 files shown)


guoqing-noaa:
Tag @mgduda for awareness. Thanks!

weiwangncar:
@guoqing-noaa Can you let us know which dataset this PR will help? I assume this will have no effect on fields that are less than 2 GB?

guoqing-noaa:
> @guoqing-noaa Can you let us know which dataset this PR will help? I assume this will have no effect on fields that are less than 2 GB?

@weiwangncar Yes, it has no effect on files less than 2GB in size. I did tests using ungrib.exe before and after this PR and I got identical intermediate binary files.

The RRFSv1 grib2 files cover the North American domain and each file is about 9GB in file size.

Also, this WRF forum post has requested the capability to handle files larger than 2GB: https://forum.mmm.ucar.edu/threads/apparent-2gb-grib-file-size-limit.8758/

Thanks!

weiwangncar:
@guoqing-noaa Thanks for the info, and that's good to know. We will run some checks too.

weiwangncar:
@jimbresch Would you be interested in taking a look at this?

PR #292, opened 2026-07-01
Address floating-point comparison issue for glevel1 and glevel2 in rd_grib2

mgduda:
This PR addresses a floating-point comparison issue for the variables `glevel1` and `glevel2` in the `rd_grib2` subroutine.

In the `rd_grib2` routine, soil levels in centimeters are computed as real values (`glevel1` and `glevel2`) from level information in GRIB files, and these are later compared with integer levels requested in the `Vtable`. Under certain compilation environments, the computed soil levels will not compare exactly with the requested integer levels, leading to the omission of soil fields from the intermediate output files.

This PR addresses the floating-point comparison issue by rounding `glevel1` and `glevel2` to the nearest integer using the Fortran `nint` intrinsic before comparing with the requested integer levels from the `Vtable`.

### DIFF
--- a/ungrib/src/rd_grib2.F
+++ b/ungrib/src/rd_grib2.F
@@ -736,8 +736,8 @@ C  SET ARGUMENTS
                 TMP8LOOP: do j = 1, maxvar
                   if ((g2code(4,j) .eq. 106) .and.
      &               (gfld%ipdtmpl(2) .eq. g2code(3,j)) .and.
-     &               (glevel1 .eq. level1(j)) .and.
-     &               ((glevel2 .eq. level2(j)) .or.
+     &               (nint(glevel1) .eq. level1(j)) .and.
+     &               ((nint(glevel2) .eq. level2(j)) .or.
      &                                   (level2(j) .le. -88))) then
                     my_field = namvar(j)
                     exit TMP8LOOP


mgduda:
@weiwangncar If you'd like to reproduce the issue that this PR aims to address, you can compile `ungrib` with the `intel/2025.3.2` module on Derecho.
```
module load intel/2025.3.2
module load ncarcompilers
git clone https://github.com/wrf-model/WPS.git
cd WPS
./configure --nowrf --build-grib2-libs
[choose option 9]
./compile ungrib
```
I was testing with a GFS GRIB2 file, but my guess is that other GRIB data may trigger the same issue.

mgduda:
@weiwangncar @jimbresch @islas This PR provides an alternative fix to the one in PR #275. The key difference is that in PR #275, the real-valued soil levels are truncated to integer values in the assignment to `glevel1` and `glevel2`, whereas this PR rounds the real values to their nearest integer values before comparing with integer levels.

weiwangncar:
@mgduda Just did a test. This PR fixes the problem with missing soil fields. I compiled the code with the current default Intel oneapi compiler version 2025.2.1. Using the original code, I get prints like these:

```
Path to intermediate files is ./
Subsoil level      0 in the GRIB2 file, was not found in the Vtable
Subsoil level      0 in the GRIB2 file, was not found in the Vtable
Subsoil level      0 in the GRIB2 file, was not found in the Vtable
Subsoil level      0 in the GRIB2 file, was not found in the Vtable
Subsoil level      0 in the GRIB2 file, was not found in the Vtable
```
With the fix in this PR, they are gone.

weiwangncar (approved):

